Issue #45, opened 2021-06-18
Equivalent of `try_files` in nginx for Angular apps


richbayliss:
# Problem
I was looking to use this as the HTTP server component in an Angular SPA but can't due to the routing mechanism in Angular.

# Question
Would you be welcome to the addition of a feature to `try_files` so that Angular SPA's might be served?

# Suggestion
Add support via a CLI argument, and code implementation like so:

```
--try-files "$uri $uri/ index.html"
```

...where `$uri` represents the requested path and is regex-replaced when finding the file to serve.

richbayliss:
I am not a massive Rust guy (yet) but here's a PR to add such a feature: https://github.com/joseluisq/static-web-server/pull/46

joseluisq:
I think that feature is not a bad idea at all if one wants to fallback on purpose to a specific html page.

joseluisq:
Hi,
I was thinking around the idea of this feature and I have concluded that it will be not accurate solution for this server.
Here my explanation why.

Nginx `try_files` directive in general is more complex that just trying a list of uris one by one until the fallback page. Which implies overhead during look up depending on the list. (I have already mentioned something about it in your PR #46)

Nginx `try_files` has different context of usage and makes sense there (For example with: location directive, proxy directive, etc). But in `static-web-server` project either `$uri` (whatever uri) and `$uri/` (whatever dir) are implicit, meaning that the server is managing those by default depending of the user request. So for example a `"try_files $uri $uri/ =404"` equivalent is redundant.

However the only thing which I see as a possibility is the fallback uri (the last parameter of `try_files`). Since this issue wants for example to redirect all request uris to another specific html page uri.
But there is a caveat with it. For example, one could introduce another flag option like `--fallback-location=/index.html` or whatever to indicate the purpose. But this could be a *temporary workaround* but I don't see it as a long-term solution since users could want to be able to fallback more than one page with not necessarily the same uri.

Nonetheless, what I could propose is to add support for _**customizable routes**_.
With this option users could add whatever routes they want to map their pages uris more precisely. But this implies necessarily to implement support for a custom config file. Something that I see factible in near future by the way.

If in the eventually case we would add routes support, then a `--fallback-location` flag option doesn't make sense or even worse if we add the fallback flag first then will make annoying (for users) to deprecate it in favor of _customizable routes_ for example.

So It's not because it's a bad idea something like `try_files` but _customizable routes_ might be even better in different aspects.

[...]

PR #46, opened 2021-06-18
✨: Support --try-files for working with SPA such as Angular

richbayliss:
Change-type: patch
Signed-off-by: Rich Bayliss <richbayliss@gmail.com>

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -12,6 +12,7 @@ pub struct RequestHandlerOpts {
     pub dir_listing: bool,
     pub cors: Option<Arc<cors::Configured>>,
     pub security_headers: bool,
+    pub try_files: Vec<String>,
 }
 
 // It defines the main request handler for Hyper service request.
@@ -30,6 +31,7 @@ impl RequestHandler {
         let root_dir = self.opts.root_dir.as_path();
         let uri_path = req.uri().path();
         let dir_listing = self.opts.dir_listing;
+        let try_files = self.opts.try_files.clone();
 
         async move {
             // CORS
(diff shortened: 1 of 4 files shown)


richbayliss:
See #45 for the motivation.

src/handler.rs
```diff
@@ -47,28 +49,37 @@ impl RequestHandler {
...
-            {
-                Ok(mut resp) => {
-                    // Append Security Headers
-                    if self.opts.security_headers {
-                        security_headers::with_security_headers(&mut resp);
-                    }
+            let mut response = error_page::get_error_response(method, &http::StatusCode::from_u16(404).unwrap());
+            for try_file in try_files.into_iter() {
+                let search_uri_path = try_file.replace("$uri", uri_path);
+                match static_files::handle_request(method, headers, root_dir, &search_uri_path, dir_listing)
```
joseluisq, line 55: Unfortunately the code is not the way to go. It has some inconsistencies.

For example there is no guarantee to pass empty values or deformed ones. Also the `search_uri_path` should be an valid URI path format but looks like is just passing the value around based on whatever string that the user can give then passed directly to `handle_request` function, which is wrong.

Other issue is that your code never evaluate for the `--try-files` as optional, the code is just assuming it as required.

Final observation, the possibly `try_files` feature could be better implemented in the `static_files` module as an enhancement for it rather than in the `handler` file over a loop.

joseluisq (changes requested):
Please improve your code and feel free to propose an alternative in order to give it a review.

joseluisq:
As I explained in issue #45 we won't add support for `try_files` due to technical reasons.
However thanks for take the time to work on this.

PR #52, opened 2021-09-06
Remove extra Docker Scratch/Alpine volumes

### LINKED ISSUE #51: Issue with volume in unRAID
First off, I don't know if it's an unRAID, Docker or the static-web-server issue. I'll post this issue in unRAID forum too.

The issue is that the static-web-server uses the `VOLUME ["/public"]` and I bind some paths into the container from unRAID which fails to umount when stopping the array. In unRAID it's called path as for the `-v /somedir:/public` argument and array manages the HDD mounting and shares. The shares in unRAID are located in `/mnt/user/<sharename>`.

I've bind 2 shared folder from unRAID into a subfolder in `/public`:
 - `/mnt/user/movies` to `/public/movies`
 - `/mnt/user/series` to `/public/series`

It's working fine so far until I want to shutdown the server or stop the array.
unRAID does these following steps:
 - stop containers graceful
 - stop docker
 - umount all shares
 - stop the array

Docker stops fine but umounting all the shares fails here because static-web-server uses `VOLUME ["/public"]` which creates a new volume by docker into `/var/lib/docker/volumes/<volume>/_data` and the both binds of `/public/movies` and `/public/series` are mounted into this folder. When docker stops the binds are not linked anymore still unRAID sees the binding between them and cannot umount the shares. To do the graceful shutdown I manually umounting both of the binds directly in the `/var/lib/docker/volumes/<volume>/_data_/movies`, kill `/dev/loop2` then unRAID is able to umount all `/mnt/user` shares.
https://github.com/joseluisq/static-web-server/blob/d165213f375b5394736c97d68b348818d79b2d4a/docker/scratch/Dockerfile#L18

If static-web-server wouldn't use `VOLUME ["/public"]` thus wouldn't create a volume under `/var/lib/docker/volumes/<volume>` and the bind goes directly into the docker process so that after stopping of docker no dead mounting exists.

joseluisq:
## Description

Scratch/Alpine Docker images create an extra Docker volume at `/public`.
The problem is that for example, an extra Docker volume can difficult umounting shares on unRAID and other environments.

## Related Issue

Resolves #51

## Motivation and Context

Since these extra Docker volumes can be optional. This PR remove them of `scratch` and `alpine` images.
So it allows to avoid extra Docker volumes for some directory (`/public`) that can be perfectly within the container. Leaving any volumes decision up to users.
For more context see #51

## How Has This Been Tested?

Using two new development Docker images. Testing should be done by @joseluisq and @bergi9

Docker development images:

- `joseluisq/static-web-server:devel` (scratch)
- `joseluisq/static-web-server:devel-alpine` (alpine)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,10 +117,18 @@ fmt:
 	@cargo fmt --all
 .PHONY: fmt
 
+docker.image:
+	@echo "Creating Docker Scratch image..."
+	@docker build \
+		--rm=true -f ./docker/scratch/Dockerfile \
+		--build-arg SERVER_VERSION="${PKG_TAG}" -t joseluisq/${PKG_NAME}:devel . --pull=true
+.PHONY: docker.image
+
 docker.image.alpine:
+	@echo "Creating Docker Alpine image..."
 	@docker build \
 		--rm=true -f ./docker/alpine/Dockerfile \
-		--build-arg SERVER_VERSION="alpine" -t ${PKG_NAME}:alpine . --pull=true
+		--build-arg SERVER_VERSION="${PKG_TAG}" -t joseluisq/${PKG_NAME}:devel-alpine . --pull=true
 .PHONY: docker.image.alpine
 
 
--- a/docker/alpine/Dockerfile
+++ b/docker/alpine/Dockerfile
@@ -19,7 +19,6 @@ COPY ./docker/alpine/entrypoint.sh /
 COPY ./docker/public /public
 
 EXPOSE 80
-VOLUME ["/public"]
 ENTRYPOINT ["/entrypoint.sh"]
 CMD ["static-web-server"]
 
--- a/docker/scratch/Dockerfile
+++ b/docker/scratch/Dockerfile
@@ -15,7 +15,6 @@ COPY --from=latest /usr/local/bin/static-web-server /
 COPY ./docker/public /public
 
 EXPOSE 80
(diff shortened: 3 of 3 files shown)


joseluisq:
Docker development images published.

- `joseluisq/static-web-server:devel` (scratch)
- `joseluisq/static-web-server:devel-alpine` (alpine)

Both with no Docker volume instructions.

@bergi9 please feel free to give them (both) a test and share your results here.

bergi9:
It doesn't create volumes anymore. Used the `joseluisq/static-web-server:devel` version. Will check it by rebooting the server tomorrow morning if it umounts correctly.

joseluisq:
Great.
Don't worry do it when you have time.
Just let me know.

I'm doing my tests on my docker environments as well for extra verification.

joseluisq:
`devel` and `devel-alpine` tested on my Linux Docker environments with no issues.

bergi9:
I rebooted the server today morning. No umount issues anymore. So I can call it as solved.

joseluisq:
Great!

BTW the next `static-web-server` release will take some time so until then feel free to continue using temporarily the `devel` or `devel-alpine` images. Both are using the latest stable `v2.0.2` binary.

joseluisq:
Released on [v2.0.3](https://github.com/joseluisq/static-web-server/releases/tag/v2.0.3)

PR #84, opened 2022-02-21
feat: added application/wasm to compression file list

acelot:
`application/wasm` added as a plain text type for compression

## Motivation and Context
`application/wasm` are plain text files and can be effectively compressed by gzip or brotli to speedup their download.

## How Has This Been Tested?
No need to test.

### DIFF
--- a/docs/content/features/compression.md
+++ b/docs/content/features/compression.md
@@ -32,6 +32,7 @@ application/vnd.ms-fontobject
 font/truetype
 font/opentype
 image/svg+xml
+application/wasm
 ```
 
 This feature is enabled by default and can be controlled by the boolean `-x, --compression` option or the equivalent [SERVER_COMPRESSION](./../configuration/environment-variables.md#server_compression) env.
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -18,7 +18,7 @@ use tokio_util::io::{ReaderStream, StreamReader};
 use crate::Result;
 
 /// Contains a fixed list of common text-based MIME types in order to apply compression.
-pub const TEXT_MIME_TYPES: [&str; 23] = [
+pub const TEXT_MIME_TYPES: [&str; 24] = [
     "text/html",
     "text/css",
     "text/javascript",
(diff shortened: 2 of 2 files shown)


joseluisq:
No bad idea.
Just curious. Can be `application/wasm` used only for text format or also covers the assembly one?

acelot:
@joseluisq Do you mean that `application/wasm` mime type can be used for compiled binary files?

acelot:
Here https://rustwasm.github.io/book/game-of-life/code-size.html we can see that `.wasm` files are typically compressed by gzip.

joseluisq:
> @joseluisq Do you mean that `application/wasm` mime type can be used for compiled binary files?

Yes, I meant If `application/wasm` can be used either with text files as well as the binary format. I just wanted to know.

acelot:
Снимок экрана 2022-02-21 в 21 47 07
128K wasm file compressed by gzip down to 50K bytes.

joseluisq:
> Here https://rustwasm.github.io/book/game-of-life/code-size.html we can see that `.wasm` files are typically compressed by gzip.

I just asked because [this](https://developer.mozilla.org/en-US/docs/WebAssembly/Text_format_to_wasm) says that `.wasm` are the binaries and `.wat` are the text ones.
Anyway it looks like both can be mime-typed as `application/wasm`.

acelot:
https://rustwasm.github.io/book/reference/code-size.html

Citing:
> WebAssembly is typically served to users gzip'd so you'll want to be sure to compare differences in gzip'd size for transfer times over the wire. Also keep in mind that the WebAssembly binary format is quite amenable to gzip compression, often getting over 50% reductions in size.

acelot:
>I just asked because [this](https://developer.mozilla.org/en-US/docs/WebAssembly/Text_format_to_wasm) says that .wasm are the binaries and .wat are the text ones.

Yeah, `.wasm` is a compiled binary files. And `.wasm` files is the only way to delivery WebAssembly to the client via browser.

joseluisq:
> https://rustwasm.github.io/book/reference/code-size.html
> 
> Citing:
> 
> > WebAssembly is typically served to users gzip'd so you'll want to be sure to compare differences in gzip'd size for transfer times over the wire. Also keep in mind that the WebAssembly binary format is quite amenable to gzip compression, often getting over 50% reductions in size.

Great! Then no doubts.

joseluisq:
Thanks!

PR #88, opened 2022-03-09
feat: add avif and JPEG XL to cache-control list.

csmith:
These are two 'next-gen' image formats, which should be cached
for the same length of time as jpegs/pngs/etc.


joseluisq (approved):

joseluisq:
Thanks!

joseluisq:
Released on [v2.7.0](https://github.com/joseluisq/static-web-server/releases/tag/v2.7.0)

PR #91, opened 2022-03-12
feat: add fallback page option

firstdorsal:
## Description
This implements a new option, `--page-fallback` / `SERVER_FALLBACK_PAGE`. If set, it is used to respond to GET requests that would otherwise result in a 404 error. Basically, it is like the `--page404` option, but returns a status code 200 instead.

## Motivation and Context
This is useful when using client routing, e.g. with react-router.

## How Has This Been Tested?
Requested a non-existing path with both the option enabled and disabled.
- enabled: the fallback page is served with a 200 status code.
- disabled: the 404 error page is served with a 404 status code.

### DIFF
--- a/src/fallback_page.rs
+++ b/src/fallback_page.rs
@@ -1,0 +1,34 @@
+use headers::{AcceptRanges, ContentLength, ContentType, HeaderMapExt, HeaderValue};
+use http::header::CONTENT_TYPE;
+use hyper::{Body, Method, Response, StatusCode};
+
+/// Checks if a fallback response can be generated, i.e. if it is a GET request that would result in a 404 error and a fallback page is configured.
+/// If a response can be generated, it is returned, else `None` is returned.
+pub fn fallback_response(
+    method: &Method,
+    status_code: &StatusCode,
+    page_fallback: &Option<String>,
+) -> Option<Response<Body>> {
+    if (status_code == &StatusCode::NOT_FOUND) && (method == Method::GET) {
+        if let Some(fallback_page_content) = page_fallback {
+            let body = Body::from(fallback_page_content.to_owned());
+            let len = fallback_page_content.len() as u64;
+
+            let mut resp = Response::new(body);
+            *resp.status_mut() = StatusCode::OK;
+            resp.headers_mut().insert(
+                CONTENT_TYPE,
+                HeaderValue::from_static("text/html; charset=utf-8"),
+            );
+            resp.headers_mut().typed_insert(ContentLength(len));
+            resp.headers_mut().typed_insert(ContentType::html());
+            resp.headers_mut().typed_insert(AcceptRanges::bytes());
+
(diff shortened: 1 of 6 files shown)


src/handler.rs
```diff
@@ -154,12 +156,19 @@ impl RequestHandler {
...
                     Ok(resp)
                 }
-                Err(status) => error_page::error_response(
-                    method,
-                    &status,
-                    &self.opts.page404,
-                    &self.opts.page50x,
-                ),
+                Err(status) => {
+                    if let Some(response) =
```
joseluisq: I would like to see the `page_fallback` option check inside the `handler.rs` rather than within the `fallback_response()`.
You can guide you from this server option.

```rs
            // `Basic` HTTP Authorization Schema
            if !self.opts.basic_auth.is_empty() {
               // ....
```

Even you could pass just a string and check if empty to skip `fallback_response()` otherwise process the `page_fallback` string value inside `fallback_response()`.

joseluisq (changes requested):

src/handler.rs
firstdorsal: I have pushed a commit to implement your changes. Please let me know if you have any further suggestions.

src/fallback_page.rs
```diff
@@ -0,0 +1,22 @@
...
+    let len = page_fallback.len() as u64;
+
+    let mut resp = Response::new(body);
+    *resp.status_mut() = StatusCode::OK;
+    resp.headers_mut().insert(
+        CONTENT_TYPE,
+        HeaderValue::from_static("text/html; charset=utf-8"),
+    );
+    resp.headers_mut().typed_insert(ContentLength(len));
+    resp.headers_mut().typed_insert(ContentType::html());
```
joseluisq: Please remove the lines starting from 13 to 16 and replace line 18 with the following:

```rs
resp.headers_mut()
        .typed_insert(ContentType::from(mime::TEXT_HTML_UTF_8));
```

Don't forget to include the `mime` crate on top via `use mime_guess::mime;`

joseluisq (changes requested):
Also since this is a new feature it needs documentation. So I suggest to document it as a sub-item (subtitle) on [error pages](https://sws.joseluisq.net/features/error-pages/).

The file is located here https://github.com/joseluisq/static-web-server/blob/master/docs/content/features/error-pages.md

And you can test it locally just running the dev server via `make docs-dev` (it requires Docker).

joseluisq:
Once changes above are done then the PR should be ready to be merged.

joseluisq (approved):

joseluisq:
Thanks!

firstdorsal:
Thank you!

joseluisq:
Released on [v2.7.0](https://github.com/joseluisq/static-web-server/releases/tag/v2.7.0)

Issue #130, opened 2022-08-11
Trailing Slash redirect and canonical URLs


joseluisq:
### Discussed in https://github.com/joseluisq/static-web-server/discussions/129

<div type='discussions-op-text'>

<sup>Originally posted by **phartenfeller** August 10, 2022</sup>
Hi!

First off thanks a lot for working on this awesome project!

I have an issue where I am not sure whether this could be a bigger SEO problem.

In #74 you implemented a feature that results in webpages (`index.html`) always being redirected to a trailing slash.

E. g. calling `.../blog/post1` will always get a 308 and redirect to `.../blog/post1/`.

I think this could cause problems in cases where websites use [canonical urls](https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls) without a trailing slash. Crawlers want to crawl the canonical document, get redirected, and not be on the canonical document. I am pretty sure that this causes errors in my google search console (error message says: redirect error). But strangely these errors do appear consistently, I have many pages indexed under the canonical URL but sometimes when I manually tell Google to index a page this error does occur. 

You could argue that I could change my canonical tags to trailing slashes on all pages. But on an established site where much is already indexed, I guess this could be a huge risk as search engines could re-index (und unrank) everything as they want the canonical document (I am absolutely not sure about this).

It is also not uncommon to use canonical URLs without trailing slashes. [Google says they don't care which to use](https://developers.google.com/search/blog/2010/04/to-slash-or-not-to-slash).

What do you think about a flag that disables the redirect logic: https://github.com/joseluisq/static-web-server/blob/6840d0fc21a4f5d742c6435987b40ecdf00f5d01/src/static_files.rs#L63-L80 . I experimented with a fork where I compiled your source code without the redirect and everything seems to work fine for me.

If you agree to a flag like this I would try to open a PR for it (I have little Rust experience).</div>

joseluisq:
Released on [v2.11.0](https://github.com/joseluisq/static-web-server/releases/tag/v2.11.0)

PR #131, opened 2022-08-13
New flag to make trailing slash redirect optional

phartenfeller:
## Description
This PR implements a new flag `redirect-trailing-slash` that is true by default resulting in no changed behavior in default settings. Setting it to false will disable the redirection to a directory path with a trailing slash (implemented in #74).

I also updated the docs to mention the new flag.

## Related Issue
Feature request #130 

## Motivation and Context
If a website has already told crawlers that the [canonical document uri](https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls) is without the trailing slash, crawlers are now unable to access it because of the permanent redirection.

## How Has This Been Tested?
I added two new tests with the new flag enabled and disabled. I needed to add a new `subdir/index.html` file in the `docker/public` directory for the tests as a request on a root `index.html` file always results in a trailing slash:

```
wget http://localhost:80
--2022-08-13 10:16:25--  http://localhost/
```

I also tested it manually by using CLI arguments, a config file, and ENV variable.

### DIFF
--- a/docker/public/subdir/index.html
+++ b/docker/public/subdir/index.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <title>Subdir Page</title>
+  <link rel="stylesheet" href="/assets/main.css">
+  <link rel="shortcut icon" href="/assets/favicon.ico">
+</head>
+
+<body>
+  <h1>Subdir Page</h1>
+  <p>A blazing fast and asynchronous web server for static files-serving. ⚡</p>
+  <p><a href="https://github.com/joseluisq/static-web-server/" target="_blank">View on GitHub</a></p>
+  <script src="/assets/main.js"></script>
+</body>
+
+</html>
--- a/docs/content/configuration/config-file.md
+++ b/docs/content/configuration/config-file.md
@@ -63,6 +63,9 @@ grace-period = 0
 #### Log request Remote Address if available
 log-remote-address = false
 
+#### Redirect to trailing slash in the requested directory uri
+redirect-trailing-slash = false
+
 
 ### Windows Only
 
--- a/docs/content/configuration/config-file.md
+++ b/docs/content/configuration/config-file.md
@@ -64,7 +64,7 @@ grace-period = 0
 log-remote-address = false
 
 #### Redirect to trailing slash in the requested directory uri
-redirect-trailing-slash = false
+redirect-trailing-slash = true
 
 
 ### Windows Only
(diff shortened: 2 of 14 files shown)


docker/public/subdir/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <title>Subdir Page</title>
```
joseluisq: I suggest moving this index file to `docker/public/assets/index.html` and removing the `subdir`.
Also, change the title to `Assets Page` one is moved.

docs/content/configuration/config-file.md
```diff
@@ -63,6 +63,9 @@ grace-period = 0
 #### Log request Remote Address if available
 log-remote-address = false
 
+#### Redirect to trailing slash in the requested directory uri
+redirect-trailing-slash = false
```
joseluisq: Since on this page, we show the default values then `redirect-trailing-slash` should be set to `true`.

src/settings/mod.rs
```diff
@@ -83,6 +83,7 @@ impl Settings {
         let mut grace_period = opts.grace_period;
         let mut page_fallback = opts.page_fallback;
         let mut log_remote_address = opts.log_remote_address;
+        let mut redirect_trailing_slash: bool = opts.redirect_trailing_slash;
```
joseluisq: No need to specify the `bool` type since it is inferred.

tests/static_files.rs
```diff
@@ -141,6 +145,53 @@ mod tests {
...
+            true,
+        )
+         .await
+            {
+                Ok(res) => {
+                    assert_eq!(res.status(), 308);
+                    assert_eq!(res.headers()["location"], "subdir/");
+                }
+                Err(status) => {
+                    panic!("expected a status 200 but not a status {}", status)
```
joseluisq: On the panic description, we are expecting `308` status.

tests/static_files.rs
```diff
@@ -141,6 +145,53 @@ mod tests {
...
+            }
+    }
+
+    #[tokio::test]
+    async fn handle_disabled_trailing_slash_redirection_subdir() {
+        match static_files::handle(
+            &Method::GET,
+            &HeaderMap::new(),
+            root_dir(),
+            "subdir",
```
joseluisq: Fix the tests changing to `assets` dir.

joseluisq (changes requested):
I will do some tests on my own but in general, after suggested changes are addressed as well as the formatting issue then the PR should be ready.

joseluisq (commented):
I just improved the options set for the static handler since the compiler complained because of too many arguments.
So you can continue with the requested changes.

phartenfeller:
Thanks for you feedback and your refactor!

I implemented the requested changes.

I guess you already resolved the formatting issues? Running `cargo fmt --all -- --check` returns nothing for me.

joseluisq (approved):

joseluisq:
Merged and ready for the next release. :+1:

PR #138, opened 2022-09-14
Add logging of real remote IP - in case of proxies

dlvoy:
Extends logging to capture `X-Forwarded-For` header (in case of proxies)  

## Description
When server is used behind reverse proxy, reported `remote_addr` reflects internal interface and port, not client external IP. This PR adds additional `real_remote_ip` that is filled with contents of X-Forwarded-For

## Motivation and Context
Main motivation is enable logs to be used with web statistics tools (like AWStats) that require IP addresses to be present in log files

## How Has This Been Tested?
Tested by providing (or not) X-Forwarded-For header with Postman

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -1,6 +1,6 @@
 use headers::HeaderValue;
 use hyper::{header::WWW_AUTHENTICATE, Body, Method, Request, Response, StatusCode};
-use std::{future::Future, net::SocketAddr, path::PathBuf, sync::Arc};
+use std::{future::Future, net::IpAddr, net::SocketAddr, path::PathBuf, sync::Arc};
 
 use crate::{
     basic_auth, compression, control_headers, cors, custom_headers, error_page, fallback_page,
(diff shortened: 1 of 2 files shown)


src/handler.rs
```diff
@@ -62,6 +62,12 @@ impl RequestHandler {
         if log_remote_addr {
             remote_addr_str.push_str(" remote_addr=");
             remote_addr_str.push_str(&remote_addr.map_or("".to_owned(), |v| v.to_string()));
+
+            let forwarded_ip = headers.get("X-Forwarded-For");
+            if let Some(client_ip_address) = forwarded_ip {
```
joseluisq:
```suggestion
            if let Some(client_ip_address) = headers.get("X-Forwarded-For") {
```

joseluisq (changes requested):
Could you just provide some log output of how this looks like?

And please add a short description about this feature on https://sws.joseluisq.net/features/logging/ and under some new `Log Real Remote IP` subtitle. See [logging.md](https://github.com/joseluisq/static-web-server/blob/master/docs/content/features/logging.md) page for more details.

dlvoy:
I've added documentation and added parsing of header - while creating documentation i found articles regarding log poisoning and simply appending contents of header would pose security risk - so current implementation requires header to contain valid, parsable IP

sample call and log to test it:

```sh
curl --header "X-Forwarded-For: 203.0.113.195, 2001:db8:85a3:8d3:1319:8a2e:370:7348" http://0.0.0.0:8080
```

```log
2022-05-23T22:24:50.519540Z  INFO static_web_server::handler: incoming request: method=GET uri=/ remote_addr=192.168.1.126:57625 real_remote_ip=203.0.113.195
```

joseluisq (approved):
Ready for the next release. Thanks :+1:

joseluisq:
Released on [v2.12.0](https://github.com/joseluisq/static-web-server/releases/tag/v2.12.0)

PR #144, opened 2022-09-24
CORS expose headers option

nelsonjchen:
When allowing headers, also expose the headers for remote scripts to use.

## Description

Allowed headers in CORS are also allowed to be used by remote scripts.

## Related Issue
Can't find any

## Motivation and Context

https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.createLazyFile depends on being able to see the values of the range and content type of the remote URL. 

I've been trying to get datasette-lite to load a 32GB SQLite FTS5 database. The CORS headers do not go far enough to satisfy `createLazyFile`. 

https://github.com/simonw/datasette-lite/pull/49 

## How Has This Been Tested?

I served a directory with:

`cargo run -- -c \* -p 3001 -z -g debug -j "origin, content-type, content-encoding, accept-ranges" -d ~/db_dir/all_records`

I was able to call it with my build from

https://github.com/simonw/datasette-lite/pull/49 

and a URL not unlike:

http://localhost:8009/?url=http%3A%2F%2Flocalhost%3A3001%2Fdb.sqlite#/db?sql=SELECT+*+FROM+records+WHERE+records+MATCH+%22hacienda+heights%22

It seemed to satisfy Enscriptem's createLazyFile from looking in the console and not throw errors complaining about not being able to access the `accept-ranges` and `content-type` headers from the code.

### DIFF
--- a/docs/content/configuration/command-line-arguments.md
+++ b/docs/content/configuration/command-line-arguments.md
@@ -41,6 +41,10 @@ OPTIONS:
     -c, --cors-allow-origins <cors-allow-origins>
             Specify an optional CORS list of allowed origin hosts separated by comas. Host ports or protocols aren't
             being checked. Use an asterisk (*) to allow any host [env: SERVER_CORS_ALLOW_ORIGINS=]  [default: ]
+        --cors-expose-headers <cors-expose-headers>
+            Specify an optional CORS list of exposed headers separated by commas. Default "origin, content-type". It
+            requires `--cors-expose-origins` to be used along with [env: SERVER_CORS_EXPOSE_HEADERS=]  [default: origin,
+            content-type]
     -z, --directory-listing <directory-listing>
             Enable directory listing for all requests ending with the slash character (‘/’) [env:
             SERVER_DIRECTORY_LISTING=]  [default: false]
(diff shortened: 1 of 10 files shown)


joseluisq:
> This is just a draft, I'm not sure why this isn't the default so I'm worried it might be for security reasons.

I would suggest that instead of forcing this to be  something by default. Why not just let users decide to expose headers when they want via a flag.

nelsonjchen:
OK, I'll see about adding this as an option.

nelsonjchen:
Properized it I guess?

nelsonjchen:
Whoops need to fix tests.

joseluisq (changes requested):
Looks fine to me.
The only thing missing is a doc entry on [CORS feature page](https://github.com/joseluisq/static-web-server/blob/master/docs/content/features/cors.md).
You can just append a new subtitle like `Expose headers` and describe a bit the feature.
The page will be published later on [sws.joseluisq.net/features/cors](https://sws.joseluisq.net/features/cors/)

joseluisq (approved):
Thanks 👍

joseluisq:
Released on [v2.13.0](https://github.com/joseluisq/static-web-server/releases/tag/v2.13.0)

PR #153, opened 2022-10-16
Make static file reading (content/metadata) operations sync

### LINKED ISSUE #146: Much slower than sfz
Just a random, single data point: I've compared req/s to https://github.com/weihanglo/sfz with a specific setup:

- Request ~ 3k files via `curl --http2 --parallel`
- Total downloaded size is ~ 6 GB

Req/s or GB/s with sfz is roughly 3x better than with static-web-server.

Using HTTP/2 doesn't make a difference.

Version: static-web-server 2.11.0

curl output of a single download:

```
sfz:
...
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< server: sfz/0.7.1
< cache-control: public, max-age=0
< last-modified: Sun, 21 Aug 2022 15:09:17 GMT
< etag: "1661094557-12909"
< accept-ranges: bytes
< content-type: application/x-bzip2
< transfer-encoding: chunked
< date: Sun, 25 Sep 2022 19:58:36 GMT

static-web-server:
...
< HTTP/1.1 200 OK
< content-length: 12909
< content-type: application/x-bzip2
< accept-ranges: bytes
< last-modified: Sun, 21 Aug 2022 15:09:17 GMT
< cache-control: public, max-age=31536000
< date: Sun, 25 Sep 2022 20:02:44 GMT
```

joseluisq:
## Description

This PR makes file content and metadata read operations _sync_ which increases performance and reduces resource utilization.

**Synopsis**

We know that the throughput is critical for file reading (content/metadata) operations on our static file module.
However, those particular operations were performed _async_, causing significant overhead due to the added extra costs of using the async runtime for them.
But, we have extremely little waiting (we essentially block on one task) for those file reading operations.
so we turned them into their _sync_ counterparts, resulting in `~58%` performance increase and `~10%` _(CPU)_ / `~52%` _(RAM)_ less resource utilization.

**Notes**

We are now using a static `4KB` (read buffer size) for Unix-like since looks like the number is _still_ the [most common page size](https://www.sobyte.net/post/2021-12/whys-the-design-linux-default-page/) for most operating systems.
For Windows, we still keep the usage of `8KB` since it looks like [it's a balanced number](https://devblogs.microsoft.com/oldnewthing/20040908-00/?p=37923).

We were inspired by the ideas of the [weihanglo/sfz's `FileStream`](https://github.com/weihanglo/sfz/blob/master/src/server/send.rs#L143-L173) implementation as analogous to our `file_stream` one.

## Related Issue

Resolves #146

## How Has This Been Tested?

Below are two 1min tests using the SWS defaults:

```sh
wrk --latency -t4 -c100 -d1m http://localhost
```

**before (v2.13.0):**

```
Running 1m test @ http://localhost
  4 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     5.47ms    1.15ms  33.78ms   75.84%
    Req/Sec     4.58k   276.47     5.74k    74.96%
  Latency Distribution
     50%    5.47ms
     75%    6.06ms
     90%    6.73ms
     99%    8.49ms
  1094168 requests in 1.00m, 747.13MB read
Requests/sec:  18222.47
Transfer/sec:     12.44MB
```

**after:**

```
Running 1m test @ http://localhost
  4 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.29ms    0.91ms  19.52ms   71.84%
    Req/Sec    10.97k   628.29    13.89k    86.83%
  Latency Distribution
     50%    2.25ms
     75%    2.82ms
     90%    3.40ms
     99%    4.73ms
  2620459 requests in 1.00m, 1.75GB read
Requests/sec:  43648.84
Transfer/sec:     29.80MB
```

No

### DIFF
--- a/src/compression_static.rs
+++ b/src/compression_static.rs
@@ -40,7 +40,7 @@ pub async fn precompressed_variant(
             filepath_precomp.display()
         );
 
-        if let Ok((meta, _)) = file_metadata(&filepath_precomp).await {
+        if let Ok((meta, _)) = file_metadata(&filepath_precomp) {
             tracing::trace!("pre-compressed file variant found, serving it directly");
 
             let encoding = if ext == "gz" { "gzip" } else { ext };
--- a/src/static_files.rs
+++ b/src/static_files.rs
@@ -1,9 +1,11 @@
-// Static File handler
-// -> Most of the file is borrowed from https://github.com/seanmonstar/warp/blob/master/src/filters/fs.rs
+//! Static File handler
+//!
+//! Part of the file is borrowed and adapted at a convenience from
+//! https://github.com/seanmonstar/warp/blob/master/src/filters/fs.rs
 
 use bytes::{Bytes, BytesMut};
-use futures_util::future::Either;
-use futures_util::{future, ready, stream, FutureExt, Stream, StreamExt};
+use futures_util::future::{Either, Future};
+use futures_util::{future, Stream};
 use headers::{
     AcceptRanges, ContentLength, ContentRange, ContentType, HeaderMap, HeaderMapExt, HeaderValue,
     IfModifiedSince, IfRange, IfUnmodifiedSince, LastModified, Range,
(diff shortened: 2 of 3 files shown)


bjornharrtell:
Very interesting! It suprises me quite a bit that rust async framework has such a large overhead.

joseluisq:
> Very interesting! It suprises me quite a bit that rust async framework has such a large overhead.

As far as I can tell, it's not only because of Tokio runtime but the `tokio::fs / async` usage in particular for our io file reads since those operations really happen `sync` under the hood (blocking system calls), and using Tokio via `tokio::fs` adds extra overhead. That's why our old `file_stream` implementation was definitely involved (as removed in the PR).

But here is a remarkable explanation from Tokio website talking about Linux `io_uring` support.

> All `tokio-uring` operations are truly `async`, **unlike APIs provided by `tokio::fs`, which run on a thread pool. Using synchronous filesystem operations from a thread pool adds significant overhead.** With `io-uring`, we can perform both network and file system operations asynchronously from the same thread. But, io-uring is a lot more.
> https://tokio.rs/blog/2021-07-tokio-uring

By the way, I have some ideas in mind to add experimental support for _really asynchronous (non-blocking) file read operations_ using `io_uring` in Linux via the [tokio-uring](https://github.com/tokio-rs/tokio-uring) crate soon.

Issue #174, opened 2023-02-22
File type detection for cache control headers is suboptimal


mfontanini:
The [code](https://github.com/static-web-server/static-web-server/blob/master/src/control_headers.rs#L20-L40) that currently detects the file type for a URI and maps it to a max age value is suboptimal. It's performing one memory allocation ("." + file type) for every defined file type. 

A better way of doing this is to first find the file type in the URI and then search for it in each file type slice. This would in turn perform 0 allocations.

joseluisq:
Resolved by #175

PR #175, opened 2023-02-22
Optimize cache control headers file type detection

mfontanini:
## Description
This optimizes the detection of file types and their mapping to their respective max ages. The upstream code is currently doing N memory allocations (the concat between "." and the extension) per request. This instead finds the extension and does a single lookup for it without performing any allocations.

## Related Issue
#174

## Motivation and Context
Something something blazingly fast?

## How Has This Been Tested?
Added a test and ensured the existing ones pass.


joseluisq (approved):
We "claim" to be [blazingly fast](https://www.youtube.com/watch?v=Z0GX2mTUtfo) so this fits perfect thanks!

PR #177, opened 2023-02-26
Minor optimizations and small improvements

### LINKED ISSUE #176: Minor optimizations / potential improvements
There's a few things that popped up while looking at the code that could be improved:
* The [`Mutex` in `FileStream`](https://github.com/static-web-server/static-web-server/blob/b2cff1b1661b071617a021a21ddc2b6c9481c5dc/src/static_files.rs#L370) is unnecessary. This was probably added as a way to get around the borrow checker but futures don't get polled concurrently so there's no need use one here.
* There's a couple of instances where `PathBuf`s are being cloned (e.g. [this one](https://github.com/static-web-server/static-web-server/blob/b2cff1b1661b071617a021a21ddc2b6c9481c5dc/src/static_files.rs#L155)) that could be avoided.
* The mime type within the content-type header is being [cloned unnecessarily](https://github.com/static-web-server/static-web-server/blob/b2cff1b1661b071617a021a21ddc2b6c9481c5dc/src/compression.rs#L73). The content-type header has a `Mime` inside which can be turned into a str ref so you can directly look it up in the slice of mime types without turning it into a `String`.

None of these are very important/should be noticeable but I think fixing them will also make the code a bit cleaner.

mfontanini:
## Description
This fixes the minor improvements brought up in #176. Namely:
* The mutex in `FileStream` is unnecessary.
* Some `PathBuf`s are being cloned unnecessarily.
* A `Mime` is being converted into a `String` when its `AsRef<str>` implementation does the trick.

## Related Issue
Closes #176

## Motivation and Context
There were no noticeable performance improvements but it's good to avoid these things (clones and mutexes) at least for the sake of code simplicity.

## How Has This Been Tested?
`cargo test` and tested in a browser.

### DIFF
--- a/src/compression_static.rs
+++ b/src/compression_static.rs
@@ -1,5 +1,6 @@
 use headers::{ContentCoding, HeaderMap, HeaderValue};
 use std::{
+    ffi::OsStr,
     fs::Metadata,
     path::{Path, PathBuf},
 };
@@ -33,11 +34,10 @@ pub async fn precompressed_variant<'a>(
 
     // Try to find the pre-compressed metadata variant for the given file path
     if let Some(ext) = precomp_ext {
-        let mut filepath_precomp = file_path.to_owned();
-        if let Some(filename) = filepath_precomp.file_name() {
-            let filename = filename.to_str().unwrap();
+        let filename = file_path.file_name().and_then(OsStr::to_str);
+        if let Some(filename) = filename {
             let precomp_file_name = [filename, ".", ext].concat();
-            filepath_precomp.set_file_name(precomp_file_name);
+            let filepath_precomp = file_path.with_file_name(precomp_file_name);
 
             tracing::trace!(
                 "getting metadata for pre-compressed file variant {}",
--- a/src/compression_static.rs
+++ b/src/compression_static.rs
@@ -1,13 +1,16 @@
 use headers::{ContentCoding, HeaderMap, HeaderValue};
-use std::{fs::Metadata, path::PathBuf};
+use std::{
+    fs::Metadata,
+    path::{Path, PathBuf},
+};
 
 use crate::{compression, static_files::file_metadata};
 
 /// Search for the pre-compressed variant of the given file path.
-pub async fn precompressed_variant(
-    file_path: PathBuf,
-    headers: &HeaderMap<HeaderValue>,
-) -> Option<(PathBuf, Metadata, &str)> {
+pub async fn precompressed_variant<'a>(
+    file_path: &Path,
+    headers: &'a HeaderMap<HeaderValue>,
+) -> Option<(PathBuf, Metadata, &'a str)> {
     let mut precompressed = None;
 
     tracing::trace!(
@@ -30,7 +33,7 @@ pub async fn precompressed_variant(
 
     // Try to find the pre-compressed metadata variant for the given file path
     if let Some(ext) = precomp_ext {
-        let mut filepath_precomp = file_path;
+        let mut filepath_precomp = file_path.to_owned();
         if let Some(filename) = filepath_precomp.file_name() {
             let filename = filename.to_str().unwrap();
             let precomp_file_name = [filename, ".", ext].concat();
--- a/src/static_files.rs
+++ b/src/static_files.rs
@@ -151,7 +151,7 @@ async fn composed_file_metadata<'a>(
     if compression_static {
         tried_precompressed = true;
         if let Some((path, meta, ext)) =
-            compression_static::precompressed_variant(file_path.clone(), headers).await
+            compression_static::precompressed_variant(file_path, headers).await
         {
             return Ok((file_path, meta, false, Some((path, ext))));
         }
@@ -179,7 +179,7 @@ async fn composed_file_metadata<'a>(
             // Second pre-compressed variant check for the given file path
             if compression_static && !tried_precompressed {
                 if let Some((path, meta, ext)) =
-                    compression_static::precompressed_variant(file_path.clone(), headers).await
+                    compression_static::precompressed_variant(file_path, headers).await
                 {
                     return Ok((file_path, meta, false, Some((path, ext))));
                 }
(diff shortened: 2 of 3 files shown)


src/compression_static.rs
```diff
@@ -30,7 +33,7 @@ pub async fn precompressed_variant(
 
     // Try to find the pre-compressed metadata variant for the given file path
     if let Some(ext) = precomp_ext {
-        let mut filepath_precomp = file_path;
+        let mut filepath_precomp = file_path.to_owned();
         if let Some(filename) = filepath_precomp.file_name() {
             let filename = filename.to_str().unwrap();
             let precomp_file_name = [filename, ".", ext].concat();
```
joseluisq: Nice!
Just looking into the code involved I realized a couple of minor improvements that we can possibly do too.

- In line 38, handling a potential panic when doing the `&str` convention.
- In line 39, reducing one allocation when extending the same UTF-8 `&str` via `String::from`. Also because `concat` is ambiguous.

I can not do the "next lines changes" suggestion directly in the GitHub review tool but feel free to incorporate/improve this below.

```rs
            if let Some(filename) = filename.to_str() {
                let mut precomp_filename = String::from(filename);
                precomp_filename.push_str(".");
                precomp_filename.push_str(ext);
                filepath_precomp.set_file_name(precomp_filename);

                tracing::trace!(
                    "getting metadata for pre-compressed file variant {}",
                    filepath_precomp.display()
                );

                if let Ok((meta, _)) = file_metadata(&filepath_precomp) {
                    tracing::trace!("pre-compressed file variant found, serving it directly");

                    let encoding = if ext == "gz" { "gzip" } else { ext };
                    precompressed = Some((filepath_precomp, meta, encoding));
                }
            }
```

src/static_files.rs
```diff
@@ -367,23 +366,15 @@ const READ_BUF_SIZE: usize = 8_192;
...
-            Ok(reader) => reader,
-            Err(err) => {
-                tracing::error!("file stream error: {:?}", err);
-                return Poll::Ready(Some(Err(anyhow::anyhow!("failed to read stream file"))));
-            }
-        };
-
         let mut buf = BytesMut::zeroed(READ_BUF_SIZE);
-        match reader.read(&mut buf[..]) {
+        match Pin::into_inner(self).reader.read(&mut buf[..]) {
```
joseluisq, line 377: Great! I definitely learned something more today.

joseluisq (dismissed):

src/compression_static.rs
joseluisq: Don't worry I just fired a commit 😅

joseluisq:
Don't worry I just fired a commit 😅
If you feel that there is more to do let me know otherwise this is just ready!

mfontanini:
Sry, I was just finishing that up. I pushed a commit that (IMO) simplifies the logic a bit more but feel free to override. The `[].concat()` should be the same as what you were suggesting, unless I'm getting it wrong. There shouldn't be extra copies in there, it will just extend the string one chunk at a time I think?

joseluisq:
> The [].concat() should be the same as what you were suggesting, unless I'm getting it wrong. There shouldn't be extra copies in there, it will just extend the string one chunk at a time I think?

As far as I'm aware `[].concat()` always allocates but `String::from` only if the input is not valid UTF-8.

But, re-thinking the matter. `String::from` does not guarantee that the conversion will be valid UTF-8 but `[].concat()` does. So we will still go with `[].concat()`.

--

From me side no other concerns, just let me know to merge.

mfontanini:
Perfect, sounds good :ok_hand:. I'm done with the changes I wanted to make so feel free to merge!

src/compression_static.rs
```diff
@@ -30,11 +34,10 @@ pub async fn precompressed_variant(
 
     // Try to find the pre-compressed metadata variant for the given file path
     if let Some(ext) = precomp_ext {
-        let mut filepath_precomp = file_path;
-        if let Some(filename) = filepath_precomp.file_name() {
-            let filename = filename.to_str().unwrap();
+        let filename = file_path.file_name().and_then(OsStr::to_str);
```
joseluisq, line 37: 👍

joseluisq:
> But, re-thinking the matter. String::from does not guarantee that the conversion will be valid UTF-8 but [].concat() does. So we will still go with [].concat().

I just correct my previous statement. Neither `String::from` nor `[].concat()` will guarantee that the string is a valid UTF-8 since those one will assume that the input string is valid UTF-8 and the other will just concatenate. Or even they could panic if it is not the case.

joseluisq:
Merged! Thanks 💯 !

mfontanini:
Awesome, thanks!

Another way to do that would be to `format!("{base}.{ext}")` but I think at that point it's more of a personal preference than anything; those should be pretty much the same.

joseluisq:
> Another way to do that would be to `format!("{base}.{ext}")` but I think at that point it's more of a personal preference than anything; those should be pretty much the same.

Yes!

Even, now that your PR is merged, I'm realizing that at line `39` we can also save another allocation via `String::from` because at this point `filename` is a valid UTF-8 because of `OsStr::to_str` ensuring that.
So `String::from` should just take a reference to the `&str` and creates a new `String` that points to the same underlying memory as the original `&str`.

https://github.com/static-web-server/static-web-server/blob/9796d35a8b50e15cbe3abecd8888065a1bdc888d/src/compression_static.rs#L37-L40

P.d. Now also knowing that `format!()` should reuse its memory when creating the resulting `String` similar to `String::from` for a valid UTF-8 input, so should be the way to go!

A bit late around here for me so feel free to submit another PR if you want to. 😅

mfontanini:
> So String::from should just take a reference to the &str and creates a new String that points to the same underlying memory as the original &str.

`String` is not a copy-on-write string, every `String` exclusively owns its memory. If you want that behavior you need to use [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html)

> Now also knowing that format!() should reuse its memory when creating the resulting String similar to String::from for a valid UTF-8 input, so should be the way to go!

Not sure I follow and I may be missing the point, but `format!` will **always** allocate given it returns a `String`. Keep in mind all of these decisions are made at compile time: unless you're using something like `Cow` that can optionally own or borrow the underlying data (and this _can_ be decided at runtime), the type will either allocate or it won't. In the case of `String` [it always does](https://doc.rust-lang.org/std/string/struct.String.html):

 > The String type is the most common string type that **has ownership** over the contents of the string. It has a close relationship with its borrowed counterpart, the primitive str.

joseluisq:
> `String` is not a copy-on-write string, every `String` exclusively owns its memory. If you want that behavior you need to use [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html)
> 
> >

Yes, you're right. Sorry about that!
I was mixing `&str`s at compile-time. Our case above is about doing a run-time conversion string.
My previous comment should be in a context like the `Cow` smart pointer.

Issue #179, opened 2023-03-03
Add support for `.html` suffix to the paths when the file doesn't exist but the .html does


syrusakbary:
Currently, certain static site generators generate the following files when generating the public folder (such as the popular next.js and others):

```
/posts.html
/posts/my-first-article.html
/posts/my-second-article.html
```

This is the case, for example, when using Next.js for generating the [Wasmer website](https://github.com/wasmerio/wasmer.io) (you can run the build locally with `npm run build` and inspect the `out` directory).

This worked great when deploying to places such as Netlify, Vercel or Cloudflare Pages. They all allow accessing those pages to the urls:

```
/posts
/posts/my-first-article
/posts/my-second-article
```

> Here are the route-matching rules in Cloudflare: https://developers.cloudflare.com/pages/platform/serving-pages/#route-matching

For example, here's the page https://wasmer.io/posts/WAI-is-the-answer which returns the content from `/posts/WAI-is-the-answer.html`.

Now, we want to move from Netlify or Cloudflare Pages to our own servers using `static-web-server`, but currently `https://localhost:9000/posts` or `https://localhost:9000/posts/WAI-is-the-answer` return 404 in the Static Web Server.

## Proposed solution

The proposed solution would be, in case `/posts` is not found (nor `/posts/index.html`) try to resolve `/posts.html`.

syrusakbary:
This can be closed as #180 is now merged in master :)

PR #180, opened 2023-03-03
Add support for .html suffixing when the path doesn't exist

syrusakbary:
This PR fixes the issue #179.

I tried to implement in multiple ways and places, but this resulted the cleaner one

### DIFF
--- a/src/static_files.rs
+++ b/src/static_files.rs
@@ -201,6 +201,9 @@ async fn composed_file_metadata<'a>(
                     if let Some(new_meta) = new_meta {
                         meta = new_meta;
                         is_dir = false;
+                    } else {
+                        // We do it to preserve previous behavior
+                        file_path.push("index.html");
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/static_files.rs
```diff
@@ -138,11 +138,34 @@ pub async fn handle<'a>(opts: &HandleOpts<'a>) -> Result<(Response<Body>, bool),
     Ok((resp, is_precompressed))
 }
 
+/// Returns the result of trying to append `.html` to the file path.
+/// * If the prefixed html path exists, it mutates the path to the prefixed one and returns the Metadata
+/// * If the prefixed html path doesn't exist, it reverts the path to it's original value
+fn prefix_file_html_metadata<'a>(
+    file_path: &'a mut PathBuf,
+) -> (&'a mut PathBuf, Option<Metadata>) {
```
joseluisq: We can elide the `'a` lifetime here.

joseluisq (changes requested):

joseluisq (approved):
Thanks!

PR #181, opened 2023-03-03
Add support for max_blocking_threads

syrusakbary:
This PR adds support for `max_blocking_threads` setting in tokio which helps on blocking and race conditions.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -13,7 +13,8 @@ use crate::{service::RouterService, Context, Result};
 /// Define a multi-thread HTTP or HTTP/2 web server.
 pub struct Server {
     opts: Settings,
-    threads: usize,
+    worker_threads: usize,
+    max_blocking_threads: usize,
 }
 
 impl Server {
@@ -24,12 +25,13 @@ impl Server {
 
         // Configure number of worker threads
         let cpus = num_cpus::get();
-        let threads = match opts.general.threads_multiplier {
+        let worker_threads = match opts.general.threads_multiplier {
             0 | 1 => cpus,
             n => cpus * n,
         };
+        let max_blocking_threads = opts.general.max_blocking_threads;
 
-        Ok(Server { opts, threads })
+        Ok(Server { opts, worker_threads, max_blocking_threads })
     }
 
     /// Build and run the multi-thread `Server` as standalone.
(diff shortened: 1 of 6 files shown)


src/server.rs
```diff
@@ -54,10 +56,11 @@ impl Server {
     where
         F: FnOnce(),
     {
-        tracing::debug!("initializing tokio runtime with multi thread scheduler");
-
+        tracing::debug!(%self.worker_threads, "initializing tokio runtime with multi thread scheduler");
+
```
joseluisq: Fix please some formatting issues.

src/settings/cli.rs
```diff
@@ -35,7 +35,7 @@ pub struct General {
     #[structopt(
         long,
         short = "n",
-        default_value = "1",
+        default_value = "4",
         env = "SERVER_THREADS_MULTIPLIER"
```
joseluisq: I'm wondering why now we should default to `4` workers instead of `1` per core (default behavior).

src/settings/cli.rs
```diff
@@ -44,6 +44,15 @@ pub struct General {
...
     pub threads_multiplier: usize,
 
+    #[structopt(
+        long,
+        short = "b",
+        default_value = "20",
+        env = "SERVER_MAX_BLOCKING_THREADS"
+    )]
+    /// Maximum number of blocking threads 
+    pub max_blocking_threads: usize,
```
joseluisq: Should [`max_blocking_threads`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.max_blocking_threads) of `20` be an acceptable default instead of Tokio's `512`?

joseluisq (changes requested):
I just left some minor comments around.

src/settings/cli.rs
john-sharratt: As this a multiplier of the number server cores anyway a factor of one should give the best CPU efficiency (more requests for the same compute) however it will not give the best latency efficiency which will benefit with more threads per core for improved concurrency. Its a balance between the two thus it depends on what you want to tune for, maybe 2 is a good compromise? Only testing will tell

src/settings/cli.rs
john-sharratt: Ah this is because of the fact we are using `static-web-server` in WASI. spawning 512 threads in WASM is quite expensive so we capped it. Maybe its best to have two defaults, one for WASM and one for x86?

joseluisq: Ok, I got it.

> Maybe its best to have two defaults, one for WASM and one for x86?

It can be, what if we come up with some `#[cfg(wasm)]` without touching x86 (applies also to `SERVER_THREADS_MULTIPLIER`)? What do you think?

src/settings/cli.rs
joseluisq: I understand. In general, 1 per core was just a ground default. But probably we could go with something that you suggested in your [other comment](https://github.com/static-web-server/static-web-server/pull/181#discussion_r1125742586) below.

syrusakbary:
Feedback addressed @joseluisq

src/server.rs
syrusakbary: Done

joseluisq (approved):
Thanks 👍

PR #183, opened 2023-03-04
Add optional support for http2

### LINKED ISSUE #182: Set http2 as an optional (but default) feature
In some systems, the `tls` (required for `http2`) can't be easily compiled / used.

Ideally, the `http2` support would be behind a compiler flag so those systems can use `static-web-server` successfully

syrusakbary:
This PR makes http2 feature optional (but deafult), so `static-web-server` can compile in systems where tls is not supported.

Fixes #182


joseluisq (approved):
`cargo run --no-default-features -- -h` works as expected.

Thanks!

PR #185, opened 2023-03-04
Make signals optional for non-unix or non-windows systems

### LINKED ISSUE #184: Make compilation work in systems without signals
Signals are not always available in all the target systems.

Ideally, we would only use signals in the systems that support it (that is, unix and windows as they are currently the main ones supported on this repo)

syrusakbary:
This PR makes static-web-server compile in systems without signals.

Fixes #184


joseluisq (changes requested):
Looks fine. Only resolve the conflicts because of one of the previous merge.

joseluisq (approved):
Thanks!

PR #211, opened 2023-05-29
Migrate from structopt to clap v3

### LINKED ISSUE #209: Boolean options without a value (--http2) are silently ignored
**Describe the bug**

Boolean options like `--http2` and `--security-headers` only work if you add `=true` to the end, which is kinda non-standard. Seems the blame lies with [clap's peculiar decision](https://github.com/clap-rs/clap/issues/1649) to ignore decades of option-parsing precedence by default. But also it seems that clap v2.0 (which comes from structopt) does not require the user to supply the `=true` and instead will just silently ignore the flag.

**To Reproduce**

```
$ static-web-server --port 8000 --log-level=trace --security-headers --root=./ 2>&1 | grep "security headers"
2023-05-29T16:19:43.950269Z  INFO static_web_server::server: security headers: enabled=false

$ static-web-server --port 8000 --log-level=trace --security-headers=true --root=./ 2>&1 | grep "security headers"
2023-05-29T16:21:45.744096Z  INFO static_web_server::server: security headers: enabled=true

$ static-web-server --port 8000 --log-level=trace --http2 --root=./ 2>&1 | grep "security headers"
2023-05-29T16:21:04.107193Z  INFO static_web_server::server: security headers: enabled=false
(also should have crashed since no tls cert flags were provided)
```

**Expected behavior**
* `--http2` should imply "true", or `--http2` should throw an error that a value is required.

**Environment and Specs**
 - static-web-server: 2.16.0 (built from commit 0f66443)
 - OS: Linux

**Additional context**
If we migrate from structopt to clapv3 (which includes structopt), then it will log `error: The argument '--http2 <HTTP2>' requires a value but none was supplied`. But I'm struggling to implement the "--http2 implies true, no value required" behavior.

I have started down this path of migrating to clapv3 here: https://github.com/static-web-server/static-web-server/compare/master...mac-chaffee:static-web-server:clap-migration

mac-chaffee:
Essentially followed this guide: https://github.com/clap-rs/clap/blob/v3.1.6/CHANGELOG.md#migrating

## Description

* Depend on clap directly
* Update use statements
* Change "short" params to single chars
* required_if -> required_if_eq
* When referencing other args like in required_if_eq, use hyphens instead of dashes
* Last param of default_value_if is now an Option

## Related Issue
Closes #209 by throwing an error if boolean flags are not given an explicit value. 

NOTE: This change has a chance of being backwards-incompatible. Would recommend bumping the minor version at least. If users are mistakenly using a boolean flag with no value (like `--directory-listing`), this change would mean that instead of ignoring the flag, SWS would crash instead.

## How Has This Been Tested?
Local testing on macOS with these flags:

```
--http2=true --http2-tls-cert=tests/tls/local.dev_cert.pem --http2-tls-key=tests/tls/local.dev_key.pem --port=8080 --root=. --log-level=trace --security-headers[=true|false]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,6 +73,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi 0.1.19",
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


joseluisq:
> NOTE: This change has a chance of being backwards-incompatible. Would recommend bumping the minor version at least. If users are mistakenly using a boolean flag with no value (like `--directory-listing`), this change would mean that instead of ignoring the flag, SWS would crash instead.

Yes, I'm fine with exiting the program at start-up time in case of boolean flags are used without values.
However, we definitely will improve this later to allow boolean flags to be used without values.

joseluisq (approved):
Thanks!

PR #215, opened 2023-06-04
Support for boolean flags without explicit values

### LINKED ISSUE #209: Boolean options without a value (--http2) are silently ignored

joseluisq:
## Description
This PR adds support for [hybrid boolean](https://github.com/clap-rs/clap/pull/3837) flags which applies to large and short boolean flag variants.

Implicit example:

```sh
static-web-server --root public/ --compression -z
```

Explicit example (note the are no spaces in between):

```sh
static-web-server --root public/ --compression=false -z
```

Note also that this PR migrates to the latest `clap` v4.

## Related Issue

Resolves #209 (post enhancement of #211)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,6 +57,55 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is-terminal",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "41ed9a86bf92ae6580e0a31281f65a1b1d867c0cc68d5346e2ae128dddfa6a7d"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


joseluisq:
@mac-chaffee It would be great a test from your side.

mac-chaffee:
tested with --security-headers and --security-headers=false and it worked

joseluisq:
> tested with --security-headers and --security-headers=false and it worked

I also will send another PR to fix `--security-headers` not set in config.

PR #254, opened 2023-08-16
Proof of concept for Lua scripts support

joseluisq:
## Description

This is just a proof of concept for supporting Lua scripts in SWS (eventually at some point) embedding the runtime.
It serves as an excuse to have a closer look at the details and possibilities of embedding Lua as well as its costs.

It works via an advanced configuration.

```toml
#....

### Scripts
[[advanced.scripts]]
# 1 MB
memory-limit = 1
# Lua opt-in stdlib modules
std = [ "io", "os", "pkg", "debug", "coroutine" ]
# Route
source = "/dynamic/**"
# Script path
destination = "tests/scripts/dynamic.lua"
```

There are some ideas to be refined in order to improve this work in progress, so will be shared here.
In case the evaluation is successful (most likely) then we can decide to include it under an experimental flag.

Any ideas are welcome.

## Motivation and Context

My interest in Lua as a scripting utility for SWS and the discussion [#253](https://github.com/orgs/static-web-server/discussions/253).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -227,6 +227,15 @@ dependencies = [
 
 [[package]]
 name = "bstr"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba3569f383e8f1598449f1a423e72e99569137b47740b1da11ef19af3d5c3223"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
+name = "bstr"
 version = "1.6.0"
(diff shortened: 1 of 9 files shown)


joseluisq:
As per the original thread on https://github.com/orgs/static-web-server/discussions/253, we are closing this PoC as it served as exploratory work.
Concluding by not adding a scripting language feature to SWS as it should play against its purpose and increase the attack surface. See full comment on https://github.com/orgs/static-web-server/discussions/253#discussioncomment-11449177.

PR #301, opened 2024-01-09
Optional `Host` URI support for the URL Redirects feature

joseluisq:
## Description

This PR adds support for optional `Host` URI redirects via the advanced configuration.
If a `host` redirect setting is specified then SWS will attempt to match the value against the incoming URI host (request), applying the required redirect entry or ignoring it otherwise.

This feature will allow users for instance to perform `www` to `non-www` redirects or vice versa.

**Config**

```toml
[advanced]

[[advanced.redirects]]
# Optional `host` redirect to be matched against the incoming host URI
host = "127.0.0.1:4433"
# Required redirect settings
source = "/{*}"
destination = "https://localhost:4433/$1"
kind = 301
```

## Motivation and Context

Due to some users asking for host-based redirection.

## How Has This Been Tested?

Integration tests

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -163,6 +163,11 @@ impl RequestHandler {
             );
         }
 
+        let host = headers
+            .get(http::header::HOST)
+            .and_then(|v| v.to_str().ok())
+            .unwrap_or("");
+
         async move {
             // Health endpoint check
             if health_request {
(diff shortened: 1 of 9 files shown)


t-e-s-tweb:
Thank you for the quick resolution. Confirmed working.

PR #339, opened 2024-04-13
Fix build error when using specific or no Cargo compression features

### LINKED ISSUE #338: Fails to compile without compression feature

- [X] I already searched, and this issue is not a duplicate.

### Issue scope

Libary (crate)

### Describe the bug

As things are right now, static-web-server cannot be used if the `compression` feature is disabled. Both disabling compression completely and using a specific feature that only enables one compression algorithm will result in compile errors.

### How to reproduce it

1. Create a Rust project.
2. Run `cargo add static-web-server --no-default-features --features compression-gzip` from the command line.
3. Run `cargo build` from the command line.

Compilation fails due to following errors:

```
error[E0609]: no field `compression` on type `cli::General`
  --> ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/static-web-server-2.28.0/src/testing.rs:31:39
   |
31 |             compression: opts.general.compression,
   |                                       ^^^^^^^^^^^ unknown field
   |
   = note: available fields are: `host`, `port`, `fd`, `threads_multiplier`, `max_blocking_threads` ... and 19 others

error[E0609]: no field `compression_static` on type `cli::General`
  --> ~.cargo/registry/src/index.crates.io-6f17d22bba15001f/static-web-server-2.28.0/src/testing.rs:32:46
   |
32 |             compression_static: opts.general.compression_static,
   |                                              ^^^^^^^^^^^^^^^^^^ unknown field
   |
   = note: available fields are: `host`, `port`, `fd`, `threads_multiplier`, `max_blocking_threads` ... and 19 others
```

That’s because `handler::RequestHandlerOpts` has the `compression` and `compression_static` fields unconditionally, yet in `cli::General` these fields depend on the `compression` feature.

This makes it impossible to disable compression or to select individual compression algorithms in the builds – it always has to be all three of them.

### Expected behavior

The project’s dependencies should compile successfully, both when compression features are disabled and when only one compression algorithm is enabled.

### Complementary information

_No response_

### Build target

x86_64-unknown-linux-gnu

### Environment and specs

- [ ] **static-web-server:** 2.28.0
- [ ] **OS:** Fedora 39
- [ ] **Arch:** x86_64

### Additional context

_No response_

joseluisq:
## Description

This PR fixes a build issue when SWS is used as a dependency and the Cargo `compression` feature is disabled completely or just some specific compression features are enabled.

Also, if **only** one compression feature is enabled (e.g. `compression-gzip`) then the preferred encoding gets re-evaluated and assigned to that compression feature format (e.g. `gzip`) but only if it was part of the `accept-encoding` request header value.

For example, now the following scenarios will work:

```toml
#....
[dependencies]
static-web-server = { version = "2.28.0", default-features = false }
# or
static-web-server = { version = "2.28.0", default-features = false, features = ["compression-brotli"] }
```

In addition, some logs are added/improved to reflect the changes.

## Related Issue

Fixes #338

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -8,13 +8,13 @@
 
 // Part of the file is borrowed from <https://github.com/seanmonstar/warp/pull/513>*
 
-#[cfg(feature = "compression-brotli")]
+#[cfg(any(feature = "compression", feature = "compression-brotli"))]
 use async_compression::tokio::bufread::BrotliEncoder;
-#[cfg(feature = "compression-deflate")]
+#[cfg(any(feature = "compression", feature = "compression-deflate"))]
 use async_compression::tokio::bufread::DeflateEncoder;
-#[cfg(feature = "compression-gzip")]
+#[cfg(any(feature = "compression", feature = "compression-gzip"))]
 use async_compression::tokio::bufread::GzipEncoder;
-#[cfg(feature = "compression-zstd")]
+#[cfg(any(feature = "compression", feature = "compression-zstd"))]
 use async_compression::tokio::bufread::ZstdEncoder;
 
 use bytes::Bytes;
@@ -61,7 +61,7 @@ pub const TEXT_MIME_TYPES: [&str; 24] = [
 ];
 
 /// Create a wrapping handler that compresses the Body of a [`hyper::Response`]
-/// using `gzip`, `deflate`, `brotli` or `zstd` if is specified in the `Accept-Encoding` header, adding
+/// using gzip, `deflate`, `brotli` or `zstd` if is specified in the `Accept-Encoding` header, adding
 /// `content-encoding: <coding>` to the Response's [`HeaderMap`].
 /// It also provides the ability to apply compression for text-based MIME types only.
 pub fn auto(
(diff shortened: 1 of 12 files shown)


src/server.rs
```diff
@@ -217,21 +217,67 @@ impl Server {
...
+            feature = "compression-brotli",
+            feature = "compression-zstd",
+        )))]
         let compression_static = false;
-        #[cfg(feature = "compression")]
+        #[cfg(any(
+            feature = "compression",
+            feature = "compression-deflate",
+            feature = "compression-gzip",
+            feature = "compression-deflate",
```
palant, line 271: Nit: you are checking for `compression-deflate` twice here and a bunch of other places.

src/compression.rs
```diff
@@ -8,13 +8,13 @@
 
 // Part of the file is borrowed from <https://github.com/seanmonstar/warp/pull/513>*
 
-#[cfg(feature = "compression-brotli")]
+#[cfg(any(feature = "compression", feature = "compression-brotli"))]
```
palant, line 11: Nit: With the `compression` feature activating the `compression-brotli` feature, I’m fairly certain that this change (and a bunch of similar ones) is unnecessary. Unless you are doing this for readability purposes.

src/server.rs
joseluisq: Good one, yeah it has to be removed.

src/compression.rs
joseluisq: Correct, I'm doing this just for readability because using `compression` already implies the other ones.

But it could be removed without a problem.

Feel free to clean this up if you think so.

PR #343, opened 2024-04-14
Fix CORS does not work when used with Basic Auth

### LINKED ISSUE #336: CORS not working when Basic Auth enabled

- [X] I already searched, and this issue is not a duplicate.

### Issue scope

Linux (E.g. Arch Linux, Debian, Alpine, etc)

### Describe the bug

CORS (Access-Control-Allow-Origin) not in response header when basic auth is enabled.

CORS works when running:
./static-web-server --port 8787 --root ./public --cors-allow-origins "*"

CORS does not work when running:
./static-web-server --port 8787 --root ./public --cors-allow-origins "*" --basic-auth 'user:$2y$10$ga2EXn12OAxgxyzRNFy7Se4OvW1fxOAogpFqJZklO0OzBzu5iuqpG'

username: user
password: bbbb

### How to reproduce it

When running with basic auth, I am getting an error:
image

image

When without the basic auth, Access-Control-Allow-* returned OK:
image

### Expected behavior

Access-Control-Allow-* returned in the response header.

### Complementary information

I have tried this on x86_64 and arm64, both returned the same error.

### Build target

x86_64-unknown-linux-musl

### Environment and specs

- [ ] **static-web-server:** [e.g. v2.28.0]
- [ ] **OS:** [Linux Mint 21.3]
- [ ] **Arch:** [x86_64 (64-bit), ARM64 (64-bit)]
- [ ] **Docker:** [e.g. 26.0.0]
- [ ] **Client:** [Brave 1.64.116, Chromium 123.0.6312.105]

### Additional context

_No response_

ms140569:
This PR fixes #336 when CORS is used together with Basic Auth.

The `Authorization` header is now added to the `cors-allow-headers` list to work with the Basic Auth.
Also, the server will allow `OPTIONS` requests without an `Authorization` header since it is not required according to the [spec](https://fetch.spec.whatwg.org/#cors-protocol-and-credentials).

Test cases will follow in a separate PR.

### DIFF
--- a/docs/content/features/cors.md
+++ b/docs/content/features/cors.md
@@ -26,7 +26,7 @@ The server also supports a list of [CORS allowed headers](https://developer.mozi
 This feature depends on `--cors-allow-origins` to be used along with this feature. It can be controlled by the string `-j, --cors-allow-headers` option or the equivalent [SERVER_CORS_ALLOW_HEADERS](../configuration/environment-variables.md#server_cors_allow_headers) env.
 
 !!! info "Tips"
-    - The default allowed headers value is `origin, content-type`.
+    - The default allowed headers value is `origin, content-type, authorization`.
     - The server also supports [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) via the `OPTIONS` method. See [Preflighted requests in CORS](./http-methods.md#preflighted-requests-in-cors).
 
 Below is an example of how to CORS.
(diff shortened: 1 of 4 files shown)


joseluisq (approved):
For now, the PR has what we need to fix the issue.
In addition, I made some minor improvements to the `src/cors.rs` as well as added/updated a few log entries and updated the CORS page mentioning the new allowed `authorization` header.

However, as discussed. It would be great to add the necessary tests for the features used together. So feel free to add them here or let me know if you want to do it in a separate PR.

joseluisq (changes requested):
Just please update the PR description and title accordingly. Also, link the issue that you are fixing.

joseluisq (approved):
Thanks!

Issue #342, opened 2024-04-13
Split up the monolithic code blocks


palant:
### Search for duplicate feature request

- [X] I already searched, and this feature request or improvement is not a duplicate.

### Feature scope

Improve existing functionality

### Feature request related to a problem

I wondered whether it would be possible to implement some improvements to make SWS more suitable for my needs. However, pretty much all the currently supported features seem to add to the following three areas:

1. Server initialization
2. Settings management
3. Request handling

While these features are called “modules,” the approach isn’t actually modular. It’s these three monolithic code blocks instead. This complexity makes adding functionality a rather tricky affair.

### Describe the solution you'd like

I’d like actual modularization, where e.g. all functionality related to Basic Auth is contained in `basic_auth.rs` rather than added to large code blocks. A naive implementation would involve a trait like the following:

```rust
pub trait Module {
  fn priority() -> u32;
  fn server_init(&self) {}
  fn handle_request(&self) {}
}
```

The methods `server_init()` and `handle_request()` would be called for all registered modules during server initialization and request handling respectively. `priority()` would be a constant determining the order in which the modules are called (alternatively it could be specified during module registration).

I’m not sure whether this naive approach would have a notable performance impact due to Rust being unable to inline the module handlers. An alternative would be using macros to call all modules in a static fashion. That alternative has the disadvantage that it no longer allows trivially extending SWS functionality with third-party modules. For example, #97 would be a prime candidate for a third-party module.

Decentralizing settings should be more complicated and may be hard to achieve without changing config file format and CLI options. This is an optional second step however.

### Describe alternatives you've considered

Things can stay as they are of course, but the complexity will only grow.

### Build target

All targets

### Additional context

I would try to help with these changes if these are considered desirable. The modularization can be implemented as a number of small changes, this doesn’t require one huge PR to do it all at once.

### Pull requests so far

- [x] #344 
- [x] #345
- [x] #346
- [x] #348
- [x] #349
- [x] #350
- [x] #351
- [x] #353
- [x] #368
- [x] #369
- [x] #372
- [x] #373

joseluisq:
Thanks for your interest in improving SWS maintainability and reusability.

Below are some general thoughts.

**Context**

For SWS at least at the moment, modules are just functionality contained in a file that can be used when needed in the SWS scope.
[...]

PR #344, opened 2024-04-14
Move health endpoint-related code into a separate file

palant:
## Description

This creates a new module `health` responsible for the health endpoint. So far it handles the server initialization in the `init()` method and request processing in the `pre_process()` method. Ideally, handling of the health setting in CLI and config file will also be added here in future (not trivial given how parsing is tied to data structures).

## Related Issue
This is a tiny first step towards #342. 

## How Has This Been Tested?
Tests pass, and I’ve added a few new tests as well. I did not perform any functional testing at this point.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -106,33 +106,33 @@ pub struct RequestHandlerOpts {
 impl Default for RequestHandlerOpts {
     fn default() -> Self {
         Self {
-            root_dir: PathBuf::new(),
-            compression: false,
+            root_dir: PathBuf::from("./public"),
+            compression: true,
             compression_static: false,
             #[cfg(feature = "directory-listing")]
             dir_listing: false,
             #[cfg(feature = "directory-listing")]
-            dir_listing_order: 0, // name
+            dir_listing_order: 6, // unordered
             #[cfg(feature = "directory-listing")]
             dir_listing_format: DirListFmt::Html,
             cors: None,
             security_headers: false,
-            cache_control_headers: false,
-            page404: PathBuf::new(),
-            page50x: PathBuf::new(),
+            cache_control_headers: true,
+            page404: PathBuf::from("./404.html"),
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -158,6 +158,7 @@ pub mod error_page;
 #[cfg_attr(docsrs, doc(cfg(feature = "fallback-page")))]
 pub mod fallback_page;
 pub mod handler;
+pub mod health;
```
joseluisq: No need to expose this module, just make it available within the crate.

src/handler.rs
```diff
@@ -161,40 +187,25 @@ impl RequestHandler {
...
+            let mut uri_path = uri.path().to_owned();
+            let uri_query = uri.query();
+
+            // Health requests aren't logged here but in health module.
             tracing::info!(
                 "incoming request: method={} uri={}{}",
                 method,
                 uri,
                 remote_addr_str,
             );
```
joseluisq, line 208: We want to log only when the `health` feature is disabled. Can we add a conditional over here?

joseluisq (changes requested):
Just minor changes but looks fine to me overall.

src/lib.rs
palant: Sure, I wasn’t certain what the policy on exposing APIs was – currently it looks like everything is being exposed.

src/handler.rs
palant: I’m not sure, I think one of us misunderstands this. There is a conditional already, a few lines above. If `health` feature is enabled, `health::pre_process()` call will produce a result, so we go into `return result` and don’t get to this logging code. This should be good enough already?

joseluisq: You're right. I just missed the return.

joseluisq (approved):
Thanks!

PR #345, opened 2024-04-16
Move metrics endpoint-related code into a separate file

palant:
## Description

This is largely analogous to #344.

Loading logger module in lib.rs was moved up for sake of consistency: modules loaded after it will have macros like `server_info` imported implicitly, the ones loaded before won't. Further alphabetical rearranging of modules was performed by rustfmt.

## Related Issue

## How Has This Been Tested?
Tests pass, and I’ve added a few new tests as well. Functional testing on Linux didn’t show any issues.


joseluisq (approved):
Thanks!

PR #346, opened 2024-04-18
Move all of Basic authentication logic into basic_auth module

palant:
## Description
This establishes the same internal API for the `basic_auth` module as for `health` and `metrics` modules, the logic is now mostly self-contained in the module. While I hoped to avoid merge conflicts, there will be at least one merge conflict with #345 (importing `server_info` in `basic_auth.rs` will no longer be necessary).

## Related Issue

## How Has This Been Tested?
Tests have been extended, so all the relevant scenarios should be tested automatically now. Functional testing on Linux didn’t show any issues.


joseluisq (approved):
Thanks!

Issue #347, opened 2024-04-18
URL Redirects handling falls back to a redirect loop in case of error


palant:
### Search for duplicate issues

- [X] I already searched, and this issue is not a duplicate.

### Issue scope

Library (crate)

### Describe the bug

While working on #342 I’ve noticed some strange logic in redirect and rewrite handling. In case of errors during calculation of the destination URL, this code will fall back to redirecting to the original URI, effectively resulting in a redirect loop. The underlying error will not be logged. While errors here are unexpected, I don’t think that this is a sensible default.

### How to reproduce it

From what I can tell, this issue is not currently reproducible. Retrieving regex captures should not fail because we run that regex on the same URI earlier. And Aho-Corasick should not fail with the parameters we give it. These are merely assumptions however, neither `regex` nor `aho_corasick` module provide any guarantees that no additional error kinds will be added in future. So I’d rather not rely on these calls never failing.

### Expected behavior

I can see two better ways to handle this. The most correct solution would be logging the error and producing a 500 response. It might be an overkill here however.

The simpler solution would be falling back to the unmodified destination URL, without resolving the captures. Since the potential error is likely somehow related to the captures, skipping this step seems acceptable.

Given that someone might need to recognize and fix an issue, I’d still prefer the former solution.

### Build target

Built from source (specify below)

### Environment and specs

- [x] **static-web-server:** current development branch (https://github.com/static-web-server/static-web-server/commit/1c4fad200997903d31ce83e50a29b0e8aea508d0)

### Additional context

I intend to fix this while I’m messing with this code for #342 anyway.

joseluisq:
Good catch! I agree with not relying on the positive case only in general. 

> I can see two better ways to handle this. The most correct solution would be logging the error and producing a 500 response. It might be an overkill here however.

I find this solution acceptable, except for responding with 500.
If you asked me, I would vote for logging the error and keeping the redirection loop because first, the user will know what is the problem (error log) and the actual behavior (loop). Of course, assuming that this redirection loop _is_ a valid behavior for leaving it intact.

In this regard, I recall situations when one could misconfigure Nginx directives for example when handling requests and redirects and then getting a redirection loop when testing or the _Too Many Redirects_ in the browser.

Anyway, this will need some discussion to decide what is appropriate. Feel free to share your thoughts.

palant:
> I find this solution acceptable, except for responding with 500.

[...]

PR #348, opened 2024-04-18
Move all redirect handling logic into the redirects module

palant:
## Description
This establishes the same internal API for the `redirects` module as for `health` and `metrics` modules, the logic is now mostly self-contained in the module. The logic changed somewhat. On the one hand, `uri.path()` is no longer being cloned – cloning isn’t necessary unless there is an actual match. On the other hand, this addresses #347 for this code.

I would have changed `check_redirection()` signature to the following:

```rust
pub fn get_redirection<'a, 'b>(
    uri_host: &str,
    uri_path: &'b str,
    redirects_opts: &'a [Redirects],
) -> Option<(&'a Redirects, Captures<'b>)> {
```

But that’s a public function, so it would be a breaking change…

## Related Issue
This is another step towards fixing #342, also addresses one half of #347.

## How Has This Been Tested?
Existing integration tests pass, additional unit tests have been added for better test coverage.

### DIFF
--- a/src/redirects.rs
+++ b/src/redirects.rs
@@ -44,7 +44,7 @@ pub(crate) fn pre_process(
         let caps_range = 0..regex_caps.len();
         let caps = caps_range
             .clone()
-            .filter_map(|i| regex_caps.get(i).map(|s| s.as_str()))
+            .map(|i| regex_caps.get(i).map(|s| s.as_str()).unwrap_or(""))
             .collect::<Vec<&str>>();
 
         let patterns = caps_range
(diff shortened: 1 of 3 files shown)


palant:
I noticed another bug in the original code: if a capture is missing, the code will currently give Aho-Corasick fewer replacements than patterns, resulting in panick (“assertion `left == right` failed: replace_all requires a replacement for every pattern in the automaton”). I’ve changed the code to treat missing captures as empty replacements instead of removing them from the list and modified the tests to trigger this scenario.

src/redirects.rs
```diff
@@ -6,7 +6,106 @@
...
+        };
+        let dest = match ac.try_replace_all(dest, &caps) {
+            Ok(dest) => dest.to_string(),
+            Err(err) => return handle_error("failed replacing captures", err, opts, req),
+        };
+
+        tracing::debug!(
+            "url redirects glob pattern destination replaced: {:?}",
+            dest
+        );
```
joseluisq, line 75: Just a comment. The placeholder replacement logic over here is almost identical to the one for URL rewrites. So we could think of moving it to a commonplace.

joseluisq (approved):
Thanks!

src/redirects.rs
palant: Ok. I’ll be moving the rewrite logic as the next thing anyway, and I’ll see whether I can merge some of the code.

PR #349, opened 2024-04-19
Move most of CORS-related code into the cors module

palant:
## Description
This establishes the same internal API for CORS handling as for `health` and `metrics`. The downside is that CORS processing has to run twice, I don’t see a good way to share state between `pre_process()` and `post_process()` at this point. Maybe with the future middleware approach this will be possible.

I also noticed something I consider a bug: error responses such as 404 only get CORS headers if there is a fallback page. It should be possible to retrieve error responses across domains boundaries as well, so that a web app can distinguish between a 404 and 503 error for example. Let me know if you want me to fix this as part of this PR (it’s a trivial addition) or a separate issue/PR.

## Related Issue

## How Has This Been Tested?
Existing integration tests pass, additional unit tests have been added for better test coverage. Functional testing on Linux looks fine.

### DIFF
--- a/src/cors.rs
+++ b/src/cors.rs
@@ -13,7 +13,10 @@ use headers::{
     HeaderMapExt, HeaderName, HeaderValue, Origin,
 };
 use http::header;
+use hyper::{Body, Request, Response, StatusCode};
 use std::collections::HashSet;
+
+use crate::{error_page, handler::RequestHandlerOpts, Error};
 
 /// It defines CORS instance.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


joseluisq:
> I also noticed something I consider a bug: error responses such as 404 only get CORS headers if there is a fallback page. It should be possible to retrieve error responses across domains boundaries as well, so that a web app can distinguish between a 404 and 503 error for example. Let me know if you want me to fix this as part of this PR (it’s a trivial addition) or a separate issue/PR.

This sounds like a bug to me as well. Feel free to fix this in a separate PR since we could track it better and include it in the future release.

joseluisq (approved):
Thanks!

joseluisq:
> > I also noticed something I consider a bug: error responses such as 404 only get CORS headers if there is a fallback page. It should be possible to retrieve error responses across domains boundaries as well, so that a web app can distinguish between a 404 and 503 error for example. Let me know if you want me to fix this as part of this PR (it’s a trivial addition) or a separate issue/PR.
> 
> This sounds like a bug to me as well. Feel free to fix this in a separate PR since we could track it better and include it in the future release.

@palant just as a reminder to fix it.

palant:
> @palant just as a reminder to fix it.

I remember. I was waiting for another pull request to be merged because this isn’t only about CORS headers.

PR #350, opened 2024-04-19
Move most of maintenance mode logic into maintenance_mode module

palant:
## Description
This establishes the same internal API for maintenance mode as for `health` and `metrics`. While at it, I adjusted `get_response()` to make producing `body_content` more straightforward – no long a mutable variable that might or might not be overwritten but two straightforward branches. Side-effect: empty maintenance mode file will no longer result in default message. IMHO this makes sense because a) someone might actually show an empty message and b) the message about falling back to default message is only logged when the maintenance mode file doesn’t exist, not when it is empty.

I also changed `DEFAULT_BODY_CONTENT` because the old message is IMHO grammatically incorrect. It should be either “server is under maintenance” or “server is in maintenance mode” (I went with the latter).

## Related Issue

## How Has This Been Tested?
I added some rudimentary unit tests. Functional testing on Linux was also successful.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -249,12 +249,8 @@ impl RequestHandler {
             }
 
             // Maintenance Mode
-            if self.opts.maintenance_mode {
-                return maintenance_mode::get_response(
-                    method,
-                    &self.opts.maintenance_mode_status,
-                    &self.opts.maintenance_mode_file,
-                );
+            if let Some(response) = maintenance_mode::pre_process(&self.opts, req) {
+                return response;
             }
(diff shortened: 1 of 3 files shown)


joseluisq (approved):
Thanks!

joseluisq:
> Side-effect: empty maintenance mode file will no longer result in default message. IMHO this makes sense because a) someone might actually show an empty message and b) the message about falling back to default message is only logged when the maintenance mode file doesn’t exist, not when it is empty.

Fair point.

PR #351, opened 2024-04-19
Move various code related to header handling to the respective modules

palant:
## Description
This adds standard `post_process()` call to various header handling modules, similar to the internal API established in #349. Merely a tiny bit of code shuffling.

## Related Issue

## How Has This Been Tested?
Existing tests pass. Functional testing on Linux looks fine.


joseluisq (approved):
Thanks!

PR #353, opened 2024-04-20
Move all rewrite handling logic into the rewrites module

palant:
## Description
This implements the same changes to rewrites as have been implemented for redirects in #348. As suggested, I merged the identical code so that it is only present in the `redirects` module now.

There are some changes to the handler: request fields can no longer be borrowed because we are now modifying the request. While I meant to implement these changes anyway, this unfortunately introduces a merge conflict with #351.

There are also some behavior changes due to the fact that destination is always being processed as a URL now and not merely in the redirect scenario:

1. Previously, internal rewrites kept the query string unchanged. Now the query string from the destination overrides the original query string if present. This actually matches the behavior of nginx rewrites. The impact should be limited because the only place where query strings currently matter is directory listings, and existing configurations shouldn’t have destinations with query strings anyway.
2. It is now possible to do an internal rewrite to a full URL which includes a different host name. IMHO that’s a useful feature because it allows defining virtual hosts that are normally only reachable via rewrite – so you can have a subdirectory mapped to a different root (similar to the nginx `alias` directive). Configuration example:
```toml
[advanced]

[[advanced.rewrites]]
source = "/test/{**}"
destination = "http://internal.local/test/$1"

[[advanced.virtual-hosts]]
host = "internal.local"
root = "/usr/srv/test"
```
This might be worth documenting. There are caveats however. First, it’s a good idea to rewrite to the same path and merely change the host name like I’ve done above, otherwise redirects when accessing directories will be a mess (this is generally an issue with rewrites). Second, someone who knows the name of the internal virtual host will still be able to access it by mapping this host name to the server’s IP address explicitly. So one should not expect anything under that host’s root to be hidden just because rewrites only map to some of the files there.

## Related Issue
This is another step towards fixing #342, also fixes the rest of #347.

## How Has This Been Tested?
Integration tests pass, some additional unit tests have been added as well. Functional testing on Linux looks fine.

### DIFF
--- a/src/rewrites.rs
+++ b/src/rewrites.rs
@@ -88,6 +88,14 @@ fn merge_uris(orig_uri: &Uri, new_uri: &str) -> Result<Uri, Error> {
     if parts.path_and_query.is_none() {
         parts.path_and_query = orig_uri.path_and_query().cloned();
     }
+    if let Some(path_and_query) = &mut parts.path_and_query {
+        if let (None, Some(query)) = (path_and_query.query(), orig_uri.query()) {
+            *path_and_query = [path_and_query.as_str(), "?", query]
+                .into_iter()
+                .collect::<String>()
+                .parse()?;
+        }
+    }
     Ok(Uri::from_parts(parts)?)
 }
 
@@ -147,7 +155,7 @@ mod tests {
             },
             Rewrites {
                 source: Regex::new(r"/(source4)/(.*)").unwrap(),
-                destination: "http://example.net:1234/destination4/$1/$2".into(),
+                destination: "http://example.net:1234/destination4/$1?$2".into(),
                 redirect: None,
             },
         ]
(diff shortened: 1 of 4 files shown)


joseluisq:
JFYI @palant I will check the PR in the next hours but maybe you want to join us in Discord to also coordinate better reviews, tasks, time, etc.

src/rewrites.rs
```diff
@@ -6,7 +6,98 @@
...
+            if let Some(port) = req.uri().port_u16() {
+                host.push_str(&format!(":{}", port));
+            }
+            if let Ok(host) = host.parse() {
+                req.headers_mut().insert(HOST, host);
+            }
+        }
+
+        None
+    }
```
joseluisq, line 77: I agree that this new improved behavior has to be documented, even showing an example and mentioning the possible caveats described in the PR.
The page is located at `docs/content/features/url-rewrites.md`.

joseluisq (changes requested):
Looks good to me. Just a minor doc to add as mentioned in the comment.

src/rewrites.rs
palant: I tried documenting this, not sure how understandable it is.

joseluisq (approved):
Thanks!

Issue #340, opened 2024-04-13
Upgrade to hyper 1.x


palant:
### Search for duplicate feature request

- [X] I already searched, and this feature request or improvement is not a duplicate.

### Feature scope

Improve existing functionality

### Describe the solution you'd like

I’m trying to use this within another server solution, and the outdated libraries make it somewhat more complicated than necessary. Is there a reason to use hyper 0.14 rather than 1.2? If not, I may be able to create a pull request. The [changes to `Body` and `Server`](https://hyper.rs/guides/1/upgrading/) should be relevant here.

### Describe alternatives you've considered

Obviously, hyper 0.14 is still supported. I merely have to juggle different versions of `http::Request` and such within the application.

### Build target

All targets

### Pull requests so far

- [x] #354
- [ ] #375
- [ ] #377

joseluisq:
The reason why we still rely on 0.14 (maintained) and we didn't start yet migrating is because there are dependencies used by SWS no ready for Hyper v1 like our fork of the `headers` crate and probably others too (to be confirmed).

Additionally, we do not still have an estimation of the effort necessary to migrate our modules. So if that could at least be known then we could have a better idea about the implications if any and plan the migration task(s).

But certainly, we know that migrate to Hyper v1 is something that we have to do at some point, so why not to get started with it

If you have ideas or want to help let us know.

palant:
I see, I assumed this was merely a matter of finding the time. Unfortunately, I am no help at creating a migration plan – I don’t have any experience with hyper, so I can only figure things out by trying.

palant:
> like our fork of the `headers` crate

@joseluisq, what is the reason for this being a fork? It’s quite difficult to figure out the changes compared to upstream, with upstream changes being merged into it. However, it is my understanding that your fork is an improved version of https://github.com/hyperium/headers/pull/70. And that pull request is merely adding new types, not changing any of the module internals. From the look of it, these types could also be added to a separate module like `headers_ext` while keeping the base `headers` module unchanged.

palant:
I went through the dependencies. #354 should take care of the `headers` dependency, updating to 0.4 should be easy then. Other than that, `http` needs to be updated to 1.0 and `http-serde` to 2.1.0. That seems to be it, no other dependencies are tied to hyper from what I can tell.

joseluisq:
[...]

PR #354, opened 2024-04-20
Replaced fork of the headers module by an in-tree handler for the Accept-Encoding header

palant:
## Description
It isn’t necessary to fork `headers` module just to implement handling of a single header, it can be done externally. I’ve moved and simplified the relevant code from the `headers-accept-encoding` module into the tree, this should make upgrading to a current version of the `headers` module considerably simpler.

Note that the original code had issues, clippy in particular complained about quite a few things. For example, I made `ContentCoding` implement the standard `From`, `FromStr` and `Display` traits. Iterator chains were also unnecessarily complicated in a few cases.

As to parsing, I didn’t merely drop configuration of separators for reasons of simplicity. Parsing was also incorrect, it allowed quoting comma-separated values which `Accept-Encoding` header doesn’t. Also, `;q=` isn’t actually a separator, it’s a parameter. The syntax allows whitespace between `;` and `q=`, and the syntax also in principle allows for other parameters to be present. There is still an issue with the range: in principle, this code allows arbitrarily large values for q, and the values up to 65.535 will also be treated differently (everything above that value is considered the same). I didn’t consider that worth fixing.

I’ve also added `ContentCoding::ANY` (corresponds to `Accept-Encoding: *`). This is a valid value which probably needs to be considered by SWS at some point.

Tests are largely unchanged. I changed method names (`prefered_encoding` => `preferred_encoding`, `from_str` => `from`, `try_from_str` => `from_str`), added `quality_values::tests::multiple_values_whitespace` test and some tests for `ContentCoding::ANY`.

Licensing is still an open question, I didn’t look into it. `headers` module is licensed under MIT, I don’t know whether their code can be licensed under Apache-2.0 as well.

## Related Issue
This is a step towards fixing #340.

## How Has This Been Tested?
All tests pass.

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -19,7 +19,7 @@ use async_compression::tokio::bufread::ZstdEncoder;
 
 use bytes::Bytes;
 use futures_util::Stream;
-use headers::{AcceptEncoding, ContentCoding, ContentType, HeaderMap, HeaderMapExt, HeaderValue};
+use headers::{ContentType, HeaderMap, HeaderMapExt, HeaderValue};
 use hyper::{
     header::{CONTENT_ENCODING, CONTENT_LENGTH},
     Body, Method, Response,
@@ -30,7 +30,11 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 use tokio_util::io::{ReaderStream, StreamReader};
 
-use crate::{http_ext::MethodExt, Result};
+use crate::{
+    headers_ext::{AcceptEncoding, ContentCoding},
+    http_ext::MethodExt,
+    Result,
+};
 
 /// Contains a fixed list of common text-based MIME types in order to apply compression.
 pub const TEXT_MIME_TYPES: [&str; 24] = [
@@ -75,9 +79,9 @@ pub fn auto(
     }
 
     // Compress response based on Accept-Encoding header
-    if let Some(encoding) = get_prefered_encoding(headers) {
+    if let Some(encoding) = get_preferred_encoding(headers) {
         tracing::trace!(
-            "prefered encoding selected from the accept-ecoding header: {:?}",
+            "preferred encoding selected from the accept-encoding header: {:?}",
             encoding
         );
 
@@ -113,7 +117,7 @@ pub fn auto(
             return Ok(zstd(head, body.into()));
         }
 
-        tracing::trace!("no compression feature matched the prefered encoding, probably not enabled or unsupported");
+        tracing::trace!("no compression feature matched the preferred encoding, probably not enabled or unsupported");
     }
 
     Ok(resp)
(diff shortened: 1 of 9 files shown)


src/headers_ext/content_coding.rs
```diff
@@ -0,0 +1,154 @@
...
+                match coding {
+                    $(ContentCoding::$coding => HeaderValue::from_static($str),)+
+                }
+            }
+        }
+    }
+}
+
+define_content_coding! {
+    ANY; "*",
```
joseluisq: > I’ve also added `ContentCoding::ANY` (corresponds to `Accept-Encoding: *`). This is a valid value which probably needs to be considered by SWS at some point.

This seems to me like a new feature for SWS. Can we please move the code (also tests) around this out to a new PR? Just for increase clarity.

src/headers_ext/accept_encoding.rs
joseluisq, line 1: > Licensing is still an open question, I didn’t look into it. headers module is licensed under MIT, I don’t know whether their code can be licensed under Apache-2.0 as well.

Since combining MIT and APACHE 2.0 is a common pattern (e.g. SWS or Rust crates usually). I do not see the MIT code imposing significant restrictions on SWS. Both are permissive licenses after all.

What we have to do is at least add the relevant license text/notices for the three files explicitly and mention the source and author. Just make sure to keep using the SPDX headers.

For example, we could go with something like:

```
// SPDX-License-Identifier: MIT
// Original code authored by [Author Name].
// Original code sourced from [Source URL].
```

src/compression.rs
```diff
@@ -218,14 +222,14 @@ pub fn create_encoding_header(existing: Option<HeaderValue>, coding: ContentCodi
     coding.into()
 }
 
-/// Try to get the prefered `content-encoding` via the `accept-encoding` header.
+/// Try to get the preferred `content-encoding` via the `accept-encoding` header.
 #[inline(always)]
-pub fn get_prefered_encoding(headers: &HeaderMap<HeaderValue>) -> Option<ContentCoding> {
+pub fn get_preferred_encoding(headers: &HeaderMap<HeaderValue>) -> Option<ContentCoding> {
```
joseluisq, line 227: Comment: Unfortunately, this function is public. So we should mention this breaking change in the next future release.
Please do not dismiss this message for now.

src/headers_ext/accept_encoding.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+impl AcceptEncoding {
+    /// Convenience method to create an `Accept-Encoding: gzip` header
+    #[inline]
+    pub fn gzip() -> Self {
+        Self(HeaderValue::from_static("gzip").into())
+    }
+
```
joseluisq: Let's remove this unused method.

src/headers_ext/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+//! header.
+//!
+
+mod accept_encoding;
+mod content_coding;
+mod quality_value;
+
+pub use accept_encoding::AcceptEncoding;
+pub use content_coding::ContentCoding;
+pub use quality_value::QualityValue;
```
joseluisq: Let's also make those three modules and their methods internal.
Likely some other methods will have to be removed as a result.

joseluisq (changes requested):
Looks good so far, just address those minor changes.

src/headers_ext/mod.rs
palant: I made all of `headers_ext` internal. Yes, some unused code had to be removed, but also code examples – doc tests don’t work with private modules. And there have been more clippy warnings, so I’ve renamed `to_static` into `as_str` to match conventions. There was also a suggestion to rename `ContentCoding::GZIP` into `ContentCoding::Gzip`, a warning I silenced for now.

palant:
All done I think.

joseluisq (approved):
It's fine to me. I just reduced a bit more the visibility of some methods.

Issue #352, opened 2024-04-20
Running benchmarks once per year is not sufficient


palant:
### Search for duplicate feature request

- [X] I already searched, and this feature request or improvement is not a duplicate.

### Feature scope

Build target (release binary)

### Feature request related to a problem

As I’m implementing changes right now, I don’t really like guessing about performance impact. I’d like to see it if my changes cause a performance regression.

### Describe the solution you'd like

I want to try running benchmarks via a Github Action, so that we have some data for each change. Yes, the action runner isn’t a platform with reliable hardware characteristics. However, if we also measure lighttpd/nginx/apache for comparison, this might be good enough to detect at least significant performance degradations.

### Describe alternatives you've considered

It would of course be possible to get a cloud server with known hardware parameters for this task, it should produce more reliable results. But that’s probably an overkill for now.

### Build target

All targets

joseluisq:
Yes, there is also people wanting to see updates about this like https://github.com/static-web-server/benchmarks/issues/3.

Regardless, I agree that in order to improve performance and prevent regressions, we need some continuous benchmarking solution.

For example, we could set up / refactor our CI workflows and run our microbenchmarks continuously. Another alternative is a  cloud solutions like https://bencher.dev/ that are free for public repositories or any other.

But the CI itself has to be improved for example to run the microbenchmarks, use valgrind to check for memory leaks (e.g.  https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml#L273), etc.

And now that the code is getting more visibility on it's own place (modules). We could better optimize and incorporate benchmarks via criterion progressively.
I do not expect to benchmark every single peace of code of course but the hot parts or where makes sense.
BTW, there here are a few for future reference.

In the other hand, the benchmark repo is an HTTP load testing benchmark comparison that could complement the whole picture but can be also automated.

Definitely, there is room for improvement in the performance front. Feel free to propose something in a draft.

joseluisq:
@palant now that https://github.com/static-web-server/static-web-server/pull/355 can we close this?

gaby:
@joseluisq @palant Any chance of adding Caddy to the benchmarks list?

palant:
> @joseluisq @palant Any chance of adding Caddy to the benchmarks list?

Sure, I’ve created #379.

[...]

PR #355, opened 2024-04-21
Run load testing benchmarks comparison for each commit via Github Actions

palant:
## Description
This is rather crude but it works. The benchmarking results are remarkably consistent, I didn’t expect that. The downside: someone has to go and check/compare them manually. There are textual results in the action output, and there is a downloadable plot and binary data as the action artifact, but there is no overview of how it develops over time. This action also takes a while despite the caching, somewhere between 3 and 4 minutes. That’s despite only testing two things: producing a static file and listing a directory.

Vegeta’s choice of colors could be better. In the image below, the lower yellow dots belong to SWS (static file), the upper ones to Apache (directory listing).

The benchmarks clearly show SWS as being the slowest web server tested. In part this may be due to lack of sendfile and memory caching capabilities. I doubt that Apache has the latter by default however, and their directory listings are still almost twice as fast. So there is clearly work ahead.

## Related Issue
#352

## How Has This Been Tested?
I tried it out in my repository fork.

vegeta-plot


palant:
I’ve produced another benchmark for the `v2.28.0` tag, just to see whether there was a recent performance regression. No, it doesn’t look like it.
vegeta-plot
I’ve also tried benchmarking `v2.0.3` tag (here without `--features=all`, feature selection isn’t supported) which turned out considerably slower.
vegeta-plot

palant:
Enabling compression in other web servers moved their results closer to SWS:

vegeta-plot

| Software | Static file<br>[min, mean, 50, 90, 95, 99, max] | Directory listing<br>[min, mean, 50, 90, 95, 99, max] |
|----------|-------------|-------------------|
| SWS | 286.604µs, 352.64µs, 344.91µs, 388.785µs, 413.808µs, 522.805µs, 1.442ms | 565.272µs, 887.106µs, 882.876µs, 951.305µs, 974.732µs, 1.064ms, 1.753ms |
| Apache | 399.944µs, 477.982µs, 467.852µs, 519.752µs, 559.698µs, 666.654µs, 1.404ms | 515.379µs, 576.061µs, 561.651µs, 621.962µs, 694.457µs, 767.767µs, 1.537ms |
| lighttpd | 267.879µs, 342.65µs, 330.91µs, 381.633µs, 421.928µs, 565.361µs, 1.491ms | 389.277µs, 476.196µs, 455.193µs, 524.572µs, 636.127µs, 805.791µs, 2.894ms |
| nginx | 238.107µs, 310.438µs, 301.219µs, 347.77µs, 390.524µs, 486.028µs, 1.21ms | 194.036µs, 240.688µs, 227.065µs, 293.748µs, 347.816µs, 399.542µs, 971.213µs |

The nginx directory listing is remarkably fast, I might check later to see whether compression really applies there.

palant:
With the plot produced by vegeta being less than optimal, I used gnuplot to produce overview graphs like the following. It’s not that I am very happy with gnuplot (e.g. label positions above the bar charts are font-dependent) but I couldn’t find a better tool to produce this kind of graph.
overview

palant:
I actually found something better than gnuplot: [graph-cli](https://github.com/mcastorina/graph-cli). Made things way simpler, the overview graphs now look like this:
overview

joseluisq:
I like the idea. I assume that you will continue improving it.
Here are a few general thoughts.

> The benchmarks clearly show SWS as being the slowest web server tested. In part this may be due to lack of sendfile and memory caching capabilities.

I wonder if it is worth it to enable sendfile and memory caching for the other web servers in the comparison because SWS does not support them yet. Maybe we could add that once SWS gets them.

Also, it would be great to consider a benchmark graph that shows the resources used by each server.

And, this other server could be added to the list https://github.com/static-web-server/benchmarks/issues/2

palant:
> I wonder if it is worth it to enable sendfile and memory caching for the other web servers in the comparison because SWS does not support them yet. Maybe we could add that once SWS gets them.

Sendfile is no use with dynamic compression, so it’s effectively disabled already. As to memory caching, I didn’t enable it explicitly but maybe it’s enabled by default somewhere.

Frankly, I’m not sure that this is any good for comparing performance of web servers. It’s very hard to produce configurations that would be in any way comparable while also being realistic. For example, how do you compare SWS with `--features=all` to Apache or lighttpd that have almost no modules loaded? And if you load more modules, is it still a realistic scenario? But it is useful to show the baseline for SWS, e.g. that directory listings can be a lot faster. Primarily it should still indicate changes in SWS performance however.

> Also, it would be great to consider a benchmark graph that shows the resources used by each server.

Sure, as soon as I have an idea how one would measure that. :sweat_smile: 

> And, this other server could be added to the list https://github.com/static-web-server/benchmarks/issues/2

So far I focused on software that can be installed via the Ubuntu package manager. I’m definitely not going to compile lwan in this Github Action, yet I’m not even sure there is a third-party Ubuntu repositories for the pre-compiled binary (not that I am very keen on using it if it exists). Also keep in mind: benchmarking another piece of software adds at least half of minute runtime to this action, so it will delay the checks for pull requests even further.

joseluisq:
> Sendfile is no use with dynamic compression, so it’s effectively disabled already. As to memory caching, I didn’t enable it explicitly but maybe it’s enabled by default somewhere.

Ok. About memory cache. As far as I did learn, some servers like Nginx use a hybrid approach. See [Nignx's hybrid disk‑and‑memory cache strategy](https://www.nginx.com/blog/cache-placement-strategies-nginx-plus#Where-is-Cached-Data-Stored). I do not know about the others. But in any case, we can keep the feature if be found as the default for those servers.

> Frankly, I’m not sure that this is any good for comparing performance of web servers. It’s very hard to produce configurations that would be in any way comparable while also being realistic. For example, how do you compare SWS with `--features=all` to Apache or lighttpd that have almost no modules loaded? And if you load more modules, is it still a realistic scenario? But it is useful to show the baseline for SWS, e.g. that directory listings can be a lot faster. Primarily it should still indicate changes in SWS performance however.

Ok, my point was more in the direction of if SWS does not support features like let's say memory cache then at least keeping the benchmark closer to what SWS offers so the benchmark may approximate itself to something more fair in the overall comparison.
But regardless, the whole idea is to improve SWS performance rather than try to compete here.

> So far I focused on software that can be installed via the Ubuntu package manager. I’m definitely not going to compile lwan in this Github Action, yet I’m not even sure there is a third-party Ubuntu repositories for the pre-compiled binary (not that I am very keen on using it if it exists). Also keep in mind: benchmarking another piece of software adds at least half of minute runtime to this action, so it will delay the checks for pull requests even further.

I just mentioned it due to a request. But I get it. If there is no official binary available and only source then we could skip it for now.

joseluisq:
Looking at the graph https://github.com/static-web-server/static-web-server/pull/355#issuecomment-2068196430, SWS has a decent performance (file) compared to for example nginx.

Anyway, it would be great to have this new graph also generated for the current v2.28.0 to compare the differences.
I guess the point will be to be able to track the performance over time at some point.

And about the directory listing. As you said, there is definitely room for improvement.

joseluisq:
BTW, In case you wonder. The CI FreeBSD tests run every time there is a change which is not nice. So I will take care of it.

palant:
> Anyway, it would be great to have this new graph also generated for the current v2.28.0 to compare the differences.

Sure, this is v2.28.0 release:

overview

Also for comparison v2.27.0 release, the times are remarkably close:

overview

I think I will now make the test run on an actual test directory rather than SWS source directory/README. Then I can use matrix to split it up into five separate tests:

1. Directory listing
2. Small static file (dynamically compressed)
3. Small static file (precompressed)
4. Large static file (dynamically compressed)
5. Large static file (precompressed)

Each of these tests will produce their own overview. This should make it easier to read given how much directory listing and file times diverge. Tests 3 and 5 are where SWS should be at a disadvantage due to sendfile.

palant:
It’s now nine tests, meaning nine graphs. But the run times actually went down a bit because these tests are run in parallel.

The results largely show SWS performing better than Apache and roughly on par with lighttpd. nginx performs best in every test except for producing a 404 error.

lighttpd has an unusual compression implementation which complicates matters. It isn’t capable of redirecting to pre-compressed files, which is why the relevant tests have to request these files directly. Yet its dynamic compression will cache compressed files, something that cannot be disabled – so it performs great whenever large files have to be dynamically compressed because it only compresses them once. lighttpd is also the software that required me to reduce the request rate here – by the time it manages to cache the compressed file, requests tend to pile up already and to overload the server.

|overview1|overview2|overview3|
|----|----|----|
|overview4|overview5|overview6|
|overview7|overview8|overview9|

palant:
Job summary will now display [tables with the benchmarking results](https://github.com/static-web-server/static-web-server/actions/runs/8789995253/attempts/1#summary-24121020400), so one no longer needs to dig into logs or download artifacts to get an idea. Unfortunately, this only allows Markdown. I cannot see any way to upload images to embed them here.

palant:
I think this is good enough now. For reference, here is a run based on v2.28.0: https://github.com/palant/static-web-server/actions/runs/8790531711. Comparing it to a run on current development branch, significant differences always affect all web servers in pretty much the same way. So if there were performance regressions since the release, these aren’t significant enough to be measurable.

.github/workflows/perfcheck.yml
```diff
@@ -0,0 +1,205 @@
...
+      - .github/workflows/perfcheck.yml
+      - .github/workflows/config/benchmarking/**
+      - .github/workflows/testroot/**
+      - .cargo/config.toml
+      - scripts/ci/vegeta_report_to_csv.jq
+      - Cargo.lock
+      - Cargo.toml
+      - src/**
+  schedule:
+    - cron: '50 01 * * *' # Every day at 01:50 UTC
```
joseluisq: I believe it's just enough to run the benchmark per commit rather than per day again.

.github/workflows/perfcheck.yml
```diff
@@ -0,0 +1,205 @@
...
+      - .cargo/config.toml
+      - scripts/ci/vegeta_report_to_csv.jq
+      - Cargo.lock
+      - Cargo.toml
+      - src/**
+  push:
+    branches:
+    - master
+    - staging
+    - trying
```
joseluisq, line 15: Let's trigger only on `master` for now.

.github/workflows/testroot/filler01.txt
joseluisq, line 1: It would be great to see the `testroot` corpus moved to its own repo within the organization so that we just clone it before running the benchmarks rather than have it in the source code repo.
Let me know if that sounds also good to you.

.github/workflows/perfcheck.yml
```diff
@@ -0,0 +1,205 @@
...
+
+    - name: Build release binary
+      run: ${{ env.CARGO_BIN }} build --bin static-web-server -vv --release ${{ env.CARGO_FEATURES }} ${{ env.TARGET_FLAGS }}
+
+    - name: Initialize graph.csv
+      run: echo "Sample, min, mean, 50th, 90th, 95th, 99th, max" > graph.csv
+
+    - name: Run SWS benchmark
+      run: |
+        NAME=$(${{ env.TARGET_DIR }}/release/static-web-server --version | sed 's/static-web-server/sws/')
```
joseluisq: Can we add something like `sws 2.28.0 (source)` just to hint that it is built from the source when shown in the graph/table/csv? or is it too much text to display?

.github/workflows/perfcheck.yml
```diff
@@ -0,0 +1,205 @@
...
+        cat graph.csv | sed -e 's/,/|/g' -e 's/"//g' -e 's/^/|/' -e 's/$/|/' -e '2 i |-|-|-|-|-|-|-|-|' >> $GITHUB_STEP_SUMMARY
+        pipx run graph-cli --barh --bar-label --xlabel '' --ylabel 'Latency (ms)' --title "${{ matrix.title }}" --figsize 1200x700 --fontsize 10 --ytick-angle 45 --ytick-align top -o overview.png graph.csv
+
+    - name: Upload benchmark plots
+      uses: actions/upload-artifact@v4
+      with:
+        name: Benchmark plots (${{ matrix.title }})
+        path: |
+          detailed_plot.html
+          overview.png
```
joseluisq: For the artifacts produced by the CI matrix, Is there a possibility to collect them and compress them to a single file rather than display the 20 individually? I think if someone is interested in looking at the details, then can always download the zip file. What do you think?

joseluisq (changes requested):
Big step overall! we will get useful insights on the way.

I just left a few minor comments to address. Let me know if something is not clear.

.github/workflows/perfcheck.yml
joseluisq: Forget about it. I think is fine.

.github/workflows/testroot/filler01.txt
palant: My concern is that the test corpus might be out of sync with the workflow, e.g. if we try to reproduce historical results. Unless we always clone specific tags for the corpus…

.github/workflows/perfcheck.yml
palant: I think that the artifact upload added a merge feature recently, this should work here. But the ZIP file will be huge for the binary reports. I’d still keep the plots and the reports as two separate artifacts, so that the plots stay at least somewhat downloadable.

.github/workflows/perfcheck.yml
palant: You mean always adding this suffix, right? Because distinguishing a build from a release tag would be a bit complicated.

.github/workflows/testroot/filler01.txt
joseluisq: > My concern is that the test corpus might be out of sync with the workflow, e.g. if we try to reproduce historical results. Unless we always clone specific tags for the corpus…

Tags are a possibility. But I guess we could also make it configurable via GitHub actions? I mean, like triggering the ci benchmark per particular commit/tag. I even remember what I did in https://github.com/static-web-server/static-web-server/pull/286 to trigger for example a workflow via a particular issue label, and you could even pass whatever string like a commit or tag for example.

Or maybe it is too early to think about this?

.github/workflows/perfcheck.yml
joseluisq: Yes, the suffix thing. To hint to users because they could think that we are benchmarking an actual release.

.github/workflows/perfcheck.yml
palant: Merging took 2 minutes 30 seconds for the binary reports (500 MB). So I removed uploading them altogether, all the other data available should do in mostly scenarios.

joseluisq (approved):
Looks good enough for now.
Thanks!

Issue #332, opened 2024-04-06
Configuration advanced.headers is not applied when html served though 'index-files' matching


fuweichin:
### Search for duplicate issues

- [X] I already searched, and this issue is not a duplicate.

### Issue scope

Configuration (e.g. TOML)

### Describe the bug

As described by title, this seems to be a bug, as use of index file is so common. Nginx uses MIME type or scripting with `$request_filename` to match custom header rules, so it doesn't have such issue.

### How to reproduce it

1. Prepare configuration file config.toml.
```toml
# ...
directory-listing = true
index-files = "index.html"

[advanced]

[[advanced.headers]]
source = "**/*.{html,js,css,png,ico,json,xml}"
headers = { Cross-Origin-Embedder-Policy = "require-corp", Cross-Origin-Opener-Policy = "same-origin" }
```
2. Start server with command `static-web-server -w config.toml`.
3. Open a browser tab, access a directory which contains index.html, open another browser tab, access the previously matched index file with explicit path, e.g. "/index.html".
4. Compare response headers of the two requests, you may see the response of directory index-matching doesn't contain CORP headers.

### Expected behavior

When matching index files, actual requested file name SHOULD be used to match '[[advanced.headers]] souce' instead of the original request uri.

By the way, resources served though "directory-listing" MAY be assumed to have a default file name and an extension of the "directory-listing-format" option value, e.g. "directory-listing.html", thus to match rules in the "[[advanced.headers]]" section.

### Build target

x86_64-pc-windows-msvc

### Environment and specs

- [ ] **static-web-server:** 2.28.0
- [ ] **OS:** Windows 10
- [ ] **Arch:** x86_64 (64-bit)
- [ ] **Docker:**
- [ ] **Client:** Chrome 123
- [ ] Specify others

fuweichin:
By the way:  In auto-generated directory listing HTML, I find that the dash character (-) in filename is percent-encoded as URL.

```html
<a href="scroll%2Dwheel%2Dtiming%2Danalyzer/">scroll-wheel-timing-analyzer/</a>
```
The dash character is not necessary to be escaped/percent-encoded as URI path segment or as URI compoment, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI

joseluisq:
https://github.com/static-web-server/static-web-server/pull/333 will fix the issue soon.

joseluisq:
[...]

PR #357, opened 2024-04-22
Improve performance of directory listings

palant:
## Description
The benchmarks in #355 indicated that SWS directory listings are slow, so I looked into how they can be improved. I managed to improve performance of the directory listing benchmark by around 10% which is ok but not really sufficient to catch up with the competition.

The performance improvements are largely due to the way the listing is constructed. There is a large improvement by not percent-decoding the file name but using the name we already have. And constructing `table_rows` variable from an iterator also improves things.

Remarkably, the changes to `sort_file_entries()` function are barely noticeable in the benchmark. Presumably, Rust optimizer already figured out that no sorting needs to be performed for `order_code == 6` and that variables like `name` don’t need to be overwritten with the same value repeatedly.

## How Has This Been Tested?
I did functional testing on Linux.

### DIFF
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -8,7 +8,7 @@
 
 use chrono::{DateTime, Local, Utc};
 use clap::ValueEnum;
-use futures_util::{future, future::Either, FutureExt};
+use futures_util::{future, future::Either};
 use headers::{ContentLength, ContentType, HeaderMapExt};
 use humansize::FormatSize;
 use hyper::{Body, Method, Response, StatusCode};
@@ -63,7 +63,7 @@ pub fn auto_index(
     let filepath = opts.filepath;
     let parent = filepath.parent().unwrap_or(filepath);
 
-    tokio::fs::read_dir(parent).then(move |res| match res {
+    match std::fs::read_dir(parent) {
         Ok(dir_reader) => Either::Left(async move {
             let is_head = opts.method.is_head();
             match read_dir_entries(
@@ -113,7 +113,7 @@ pub fn auto_index(
             };
             Either::Right(future::err(status))
         }
-    })
+    }
 }
 
 const STYLE: &str = r#"<style>html{background-color:#fff;-moz-osx-font-smoothing:grayscale;-webkit-font-smoothing:antialiased;min-width:20rem;text-rendering:optimizeLegibility;-webkit-text-size-adjust:100%;-moz-text-size-adjust:100%;text-size-adjust:100%}:after,:before{box-sizing:border-box;}body{padding:1rem;font-family:Consolas,'Liberation Mono',Menlo,monospace;font-size:.75rem;max-width:70rem;margin:0 auto;color:#4a4a4a;font-weight:400;line-height:1.5}h1{margin:0;padding:0;font-size:1rem;line-height:1.25;margin-bottom:0.5rem;}table{width:100%;table-layout:fixed;border-spacing: 0;}hr{border-style: none;border-bottom: solid 1px gray;}table th,table td{padding:.15rem 0;white-space:nowrap;vertical-align:top}table th a,table td a{display:inline-block;white-space:nowrap;overflow:hidden;text-overflow:ellipsis;max-width:95%;vertical-align:top;}table tr:hover td{background-color:#f5f5f5}footer{padding-top:0.5rem}table tr th{text-align:left;}@media (max-width:30rem){table th:first-child{width:20rem;}}</style>"#;
@@ -143,7 +143,7 @@ struct SortingAttr<'a> {
 /// It reads a list of directory entries and create an index page content.
 /// Otherwise it returns a status error.
 async fn read_dir_entries(
-    mut dir_reader: tokio::fs::ReadDir,
+    dir_reader: std::fs::ReadDir,
     base_path: &str,
     uri_query: Option<&str>,
     is_head: bool,
@@ -155,12 +155,9 @@ async fn read_dir_entries(
     let mut files_count: usize = 0;
     let mut file_entries: Vec<FileEntry> = vec![];
 
-    while let Some(dir_entry) = dir_reader
-        .next_entry()
-        .await
-        .with_context(|| "unable to read directory entry")?
-    {
-        let meta = match dir_entry.metadata().await {
+    for dir_entry in dir_reader {
+        let dir_entry = dir_entry.with_context(|| "unable to read directory entry")?;
+        let meta = match dir_entry.metadata() {
             Ok(m) => m,
             Err(err) => {
                 tracing::error!(
@@ -217,7 +214,7 @@ async fn read_dir_entries(
                 }
             };
 
-            let symlink_meta = match tokio::fs::symlink_metadata(&symlink).await {
+            let symlink_meta = match std::fs::symlink_metadata(&symlink) {
                 Ok(v) => v,
                 Err(err) => {
                     tracing::error!(
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -14,7 +14,6 @@ use humansize::FormatSize;
 use hyper::{Body, Method, Response, StatusCode};
 use mime_guess::mime;
 use percent_encoding::{percent_decode_str, utf8_percent_encode, NON_ALPHANUMERIC};
-use std::cmp::Ordering;
 use std::future::Future;
 use std::io;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


palant:
For reference, I did try using `write!()` into a string to compose the entire page and not merely the table. This made the code easier to follow in some respects but it didn’t have any noteworthy impact on the performance. So it seems that there are no significant gains left in the string composition any more. Yes, I should probably try an actual profiler.

joseluisq:
Looking at the benchmark results, it seems like there is a slightly small improvement when listing large directory files but for small directories is even slightly slower. I did compare the current master results with the PR.

src/directory_listing.rs
```diff
@@ -480,40 +483,38 @@ fn sort_file_entries(files: &mut [FileEntry], order_code: u8) -> SortingAttr<'_>
...
-        }
-
-        // Modified (asc, desc)
-        2 => {
-            last_modified = "3";
-            a.modified.cmp(&b.modified)
+    match order_code {
+        0 | 1 => {
+            // Name (asc, desc)
+            files.sort_by_cached_key(|f| f.name.to_lowercase());
```
joseluisq, line 486: Is there a difference when going with `sort_by_cached_key` here rather than just `sort_by_key` in the scope?
Because what I get from the docs is that `sort_by_cached_key` performs better (cache) for functions with complex computations or operations but can be slower for simple property access, which is the case here.

src/directory_listing.rs
```diff
@@ -435,30 +434,34 @@ fn html_auto_index<'a>(
...
-    }
+    let table_rows = if base_path == "/" {
+        String::new()
+    } else {
+        String::from(r#"<tr><td colspan="3"><a href="../">../</a></td></tr>"#)
+    };
 
-    for entry in entries {
-        let file_name = &entry.name_encoded;
+    let table_rows = entries.iter().fold(table_rows, |mut output, entry| {
```
joseluisq, line 440: The code looks much more readable than the previous one. However, did you know if the iterator version here performs better in general than a simple for loop? Maybe is something that we have to double-check (benchmark/profile).

src/directory_listing.rs
```diff
@@ -435,30 +434,34 @@ fn html_auto_index<'a>(
...
+        let file_name = &entry.name;
+        let file_name_suffix = if entry.is_dir { "/" } else { "" };
         let file_modified = &entry.modified;
-        let file_uri = &entry.uri.clone().unwrap_or_else(|| file_name.to_owned());
-        let file_name_decoded = percent_decode_str(file_name).decode_utf8()?.to_string();
```
joseluisq, line 447: Are you sure if the `percent_decode_str` is not necessary here, especially in environments like Windows where files/dirs can likely contain special chars? I guess we also have to check if those names are displayed correctly if not yet.

joseluisq (changes requested):
I left some minor comments to check first.
But overall for me looks ok.

src/directory_listing.rs
joseluisq: Also, even an integration test case in `tests/dir_listing.rs` should be not bad.

palant:
> Looking at the benchmark results, it seems like there is a slightly small improvement when listing large directory files but for small directories is even slightly slower.

The benchmark results aren’t stable enough to measure small time differences reliably. You have to consider the other web servers as the baseline – their time also went up, so it’s something related to the action runner, not the code. If I check the proportion for SWS time and other servers on the 50th percentile (median), for an empty directory it went down consistently. Not a huge change but a few percent. In fact, the results are actually less conclusive for large directories.

src/directory_listing.rs
palant: `to_lowercase()` isn’t a simple property access, string manipulation isn’t cheap. I suspect that Rust optimizes for the scenario where the string is already all lowercase however which is the generally case for our test directory (not that we test sorted access there).

src/directory_listing.rs
palant: Iterators generally have exactly the same performance as for loops in Rust, the optimizer takes care of it. The difference here is not the loop but how the resulting string is being constructed.

src/directory_listing.rs
palant: Yes, `percent_decode_str` is definitely not necessary given that we use `entry.name` rather than `entry.name_encoded` – that call was merely reverting encoding we’ve done previously. File names not using proper UTF-8 encoding will fail earlier, in `read_dir_entries()` where we try to attempt converting paths into UTF-8 strings.

In fact, the correct approach would be using `OsString` for `entry.name` here and only performing string conversion for displaying the file name (using `to_string_lossy()` then to deal with invalid encodings). Also, `percent_encode` should be called instead of `utf8_percent_encode`. But dealing with non-UTF8 file names is something for the next PR, I suspect that the code actually loading them also has issues.

palant:
> > Looking at the benchmark results, it seems like there is a slightly small improvement when listing large directory files but for small directories is even slightly slower.
> 
> The benchmark results aren’t stable enough to measure small time differences reliably. You have to consider the other web servers as the baseline – their time also went up, so it’s something related to the action runner, not the code. If I check the proportion for SWS time and other servers on the 50th percentile (median), for an empty directory it went down consistently. Not a huge change but a few percent. In fact, the results are actually less conclusive for large directories.

Looking at the new test run – yes, the performance changes here are negligible. Somehow that looked differently in my previous benchmark runs.

src/directory_listing.rs
palant: > Also, even an integration test case in `tests/dir_listing.rs` should be not bad.

Actually, there is a test already: https://github.com/static-web-server/static-web-server/blob/ddda871871ea917d92b05741e4b1bbba7c27d42f/tests/dir_listing.rs#L214-L219

I don’t see what else I could test here, other than a file name which is non-UTF8. But while git appears to support such file names, checking in such a file might cause issues somewhere. And currently this test cannot be expected to pass given the issue I mentioned above.

palant:
> > Looking at the benchmark results, it seems like there is a slightly small improvement when listing large directory files but for small directories is even slightly slower.
> 
> The benchmark results aren’t stable enough to measure small time differences reliably. You have to consider the other web servers as the baseline – their time also went up, so it’s something related to the action runner, not the code. If I check the proportion for SWS time and other servers on the 50th percentile (median), for an empty directory it went down consistently. Not a huge change but a few percent. In fact, the results are actually less conclusive for large directories.

I’ve done some benchmarking on my computer, the numbers are more stable then. I could confirm – the performance for a large directory does improve but it is only around 3%. Not overwhelming but measurable.

src/directory_listing.rs
joseluisq: > In fact, the correct approach would be using `OsString` for `entry.name` here and only performing string conversion for displaying the file name (using `to_string_lossy()` then to deal with invalid encodings). Also, `percent_encode` should be called instead of `utf8_percent_encode`. But dealing with non-UTF8 file names is something for the next PR, I suspect that the code actually loading them also has issues.

That sounds like a good improvement then.

joseluisq: > Yes, `percent_decode_str` is definitely not necessary given that we use `entry.name` rather than `entry.name_encoded` – that call was merely reverting encoding we’ve done previously. File names not using proper UTF-8 encoding will fail earlier, in `read_dir_entries()` where we try to attempt converting paths into UTF-8 strings.

Ok, in that regard.
A user [reported](https://github.com/static-web-server/static-web-server/issues/332#issuecomment-2041646442) that we are encoding unreserved marks for the file names. For me, this is a bug so it would be perfect for another fix PR. Let me know if you want to also take care of it.

joseluisq: > > Yes, `percent_decode_str` is definitely not necessary given that we use `entry.name` rather than `entry.name_encoded` – that call was merely reverting encoding we’ve done previously. File names not using proper UTF-8 encoding will fail earlier, in `read_dir_entries()` where we try to attempt converting paths into UTF-8 strings.
> 
> Ok, in that regard. A user [reported](https://github.com/static-web-server/static-web-server/issues/332#issuecomment-2041646442) that we are encoding unreserved marks for the file names. For me, this is a bug so it would be perfect for another fix PR. Let me know if you want to also take care of it.

It is not that the whole URI is encoded but the file name.
Just correcting my previous comment on https://github.com/static-web-server/static-web-server/issues/332#issuecomment-2075858645.

joseluisq (approved):
Nice! Dir listing is now way faster than before. 
https://github.com/static-web-server/static-web-server/actions/runs/8823589340

Awesome!

PR #359, opened 2024-04-23
Apply the usual post-processing to error responses for consistency

palant:
## Description
This makes sure that the same post-processing applies to error pages as to regular responses. This leaves out the error page displayed if compression fails, I intend to fix this part with the #342 changes for compression.

## Related Issue
See https://github.com/static-web-server/static-web-server/pull/349#issuecomment-2067661389

## Motivation and Context
After thinking a bit more about this, I don’t see a reason to skip the usual post-processing for error pages. The steps in detail:

1. CORS: As already established in #349, this shouldn’t be skipped for error pages.
2. Compression: There is no harm in compressing error pages apart from slightly reduced performance, and other web servers seem to do it (granted, they also have larger default error pages). Note that currently compression is skipped because MIME type `text/html; charset=utf-8` doesn’t match the expected `text/html` – this is a bug which I intend to fix separately. IMHO this is the one step which we could explicitly skip for error pages, we should go by `resp.status()` then.
3. `Cache-Control` header: same caching strategy for errors as for other content seems appropriate.
4. Security headers: definitely necessary. Not having security headers on some of the responses is how these security measures are being circumvented.
5. Custom headers: these are typically used to custom security headers definitions, so these also should be applied consistently.

## How Has This Been Tested?
Tests pass, functional testing on Linux looks fine.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -292,7 +292,7 @@ impl RequestHandler {
                             &status,
                             &self.opts.page404,
                             &self.opts.page50x,
-                        )?
+                        )?,
                     );
 
                     (resp, false, None)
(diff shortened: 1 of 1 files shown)


joseluisq (approved):
I liked the idea of being consistent.
BTW I see those changes more as an improvement rather than a bug per se.
Thanks!

PR #360, opened 2024-04-23
Improve recognition of text MIME types for compression

palant:
## Description
If you look at the [list of MIME types potentially produced by the `mime_guess` package](https://github.com/abonander/mime_guess/blob/master/src/mime_types.rs), the list in `compression::TEXT_MIME_TYPES` isn’t really great for recognizing text MIME types. This implements the following changes:

* Generic recognition for MIME types starting with `text/` or ending in `+xml`/`+json`
* Faster testing of the remaining MIME type list by using a `HashSet`
* Recognition no longer confused by `charset` parameter (a bug I mentioned in #359)
* Adjusted font MIME types to the ones actually produced by `mime_guess` and removed the outdated `application/x-javascript` type 

Note that `lazy_static` isn’t a new dependency, we already have it as a transitive dependency via `tracing-subscriber`.

## How Has This Been Tested?
Functional testing on Linux looks fine, and Rust files (`text/x-rust`) are now being compressed as well.

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -24,8 +24,10 @@ use hyper::{
     header::{CONTENT_ENCODING, CONTENT_LENGTH},
     Body, Method, Response,
 };
-use mime_guess::Mime;
+use lazy_static::lazy_static;
+use mime_guess::{mime, Mime};
 use pin_project::pin_project;
+use std::collections::HashSet;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use tokio_util::io::{ReaderStream, StreamReader};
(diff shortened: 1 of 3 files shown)


src/compression.rs
```diff
@@ -36,33 +38,19 @@ use crate::{
     Result,
 };
 
-/// Contains a fixed list of common text-based MIME types in order to apply compression.
-pub const TEXT_MIME_TYPES: [&str; 24] = [
```
palant, line 40: Unfortunately, this was public, so this is a breaking change.

palant:
Benchmark results show a clear improvement on small text files (going down in the noise on large ones). Seemingly no real change on the directory index (`text/html` was listed first, so recognizing this MIME type was faster).

src/compression.rs
joseluisq: Let's keep the comment open for mentioning this breaking change in the next release.

src/compression.rs
```diff
@@ -36,33 +38,19 @@ use crate::{
...
+    /// Contains a fixed list of common text-based MIME types that aren't recognizable in a generic way.
+    static ref TEXT_MIME_TYPES: HashSet<&'static str> = [
+        "application/rtf",
+        "application/javascript",
+        "application/json",
+        "application/xml",
+        "font/ttf",
+        "application/font-sfnt",
+        "application/vnd.ms-fontobject",
+        "application/wasm",
```
joseluisq, line 51: Comment: the list is fine. But we could extend it a bit to other text or even some binary types in another PR.
E.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
For example, `font/woff` is missing and some others.

joseluisq (approved):
Thanks!

src/compression.rs
palant: I thought about `font/woff` already. But this format is compressed already. Most of the others are either uncommon on the web (`image/bmp`) or also compressed.

joseluisq: > I thought about `font/woff` already.

Are you sure? Because I saw that Cloudflare compresses them for example (e.g. brotli or gzip) with some kb's gains.
See https://developers.cloudflare.com/speed/optimization/content/brotli/content-compression/

Btw we could guide us from its content types list for a future improvement.

palant: I don’t know what Cloudflare does and why but WOFF is definitely compressed. From https://www.w3.org/TR/WOFF20ER/:

> Web Open Font Format (WOFF) 2.0 is a proposed update to the existing [WOFF 1.0](https://www.w3.org/TR/WOFF/) with improved compression.

In fact, WOFF 2.0 uses the same Brotli compression.

palant:
This caused a significant performance regression on directory listings for some reason.

palant:
Ah, I see – not actually a regression. Previously, we wouldn’t compress directory listings because the original logic wouldn’t recognize `text/html; encoding=utf-8` as a text format. And now that we do compress directory listings, performance naturally went down.

joseluisq:
> Ah, I see – not actually a regression. Previously, we wouldn’t compress directory listings because the original logic wouldn’t recognize `text/html; encoding=utf-8` as a text format. And now that we do compress directory listings, performance naturally went down.

It would be expected, but is there a big margin?

palant:
> It would be expected, but is there a big margin?

Sure, compression takes time (gzip in particular which is what vegeta is testing). So directory listing tests went up by almost 50%.

PR #361, opened 2024-04-23
Fix Accept-Encoding handling to work correctly if only two compression schemes are available

palant:
## Description
This fixes `compression::get_preferred_encoding()` to consider all encodings listed in `Accept-Encoding` header and not only the top one. While at it, I replaced dynamic construction of available encodings on each call by a constant.

## Motivation and Context
As things are right now, `compression::get_preferred_encoding()` only handles the scenarios “one compression scheme enabled” and “all compression schemes enabled” correctly, with the former being handled as a special case. If for example only `compression-deflate` and `compression-gzip` features are enabled and the client requests `Accept-Encoding: br, gzip`, it will return `None` because Brotli isn’t enabled – yet it should have recognized gzip being enabled.

## How Has This Been Tested?
I’ve adjusted the integration tests to explicitly test for the scenario where the top listed encoding isn’t something we support (`identity`). While at it, I fixed a bug in this test – it failed if only some compression features were enabled but not all.

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -64,6 +64,18 @@ pub const TEXT_MIME_TYPES: [&str; 24] = [
     "application/wasm",
 ];
 
+/// List of encodings that can be handled given enabled features.
+const AVAILABLE_ENCODINGS: &[ContentCoding] = &[
+    #[cfg(any(feature = "compression", feature = "compression-deflate"))]
+    ContentCoding::DEFLATE,
+    #[cfg(any(feature = "compression", feature = "compression-gzip"))]
+    ContentCoding::GZIP,
+    #[cfg(any(feature = "compression", feature = "compression-brotli"))]
+    ContentCoding::BROTLI,
+    #[cfg(any(feature = "compression", feature = "compression-zstd"))]
+    ContentCoding::ZSTD,
+];
+
 /// Create a wrapping handler that compresses the Body of a [`hyper::Response`]
 /// using gzip, `deflate`, `brotli` or `zstd` if is specified in the `Accept-Encoding` header, adding
 /// `content-encoding: <coding>` to the Response's [`HeaderMap`].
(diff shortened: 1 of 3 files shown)


src/compression.rs
```diff
@@ -228,43 +240,10 @@ pub fn get_preferred_encoding(headers: &HeaderMap<HeaderValue>) -> Option<Conten
     if let Some(ref accept_encoding) = headers.typed_get::<AcceptEncoding>() {
         tracing::trace!("request with accept-ecoding header: {:?}", accept_encoding);
```
joseluisq:
```suggestion
        tracing::trace!("request with accept-encoding header: {:?}", accept_encoding);
```

src/headers_ext/accept_encoding.rs
```diff
@@ -69,11 +63,6 @@ mod tests {
         let val = HeaderValue::from_static("deflate, gzip;q=1.0, br;q=0.9");
```
joseluisq: Let's add `zstd` to the list just for completeness.

joseluisq (changes requested):
Minor stuff to address and this bug below.

If for example, you use the SWS as a crate and enable just one compression feature like `default-features = false, features = ["compression-brotli"]` then you will get the following error.

```sh
error: variable does not need to be mutable
   --> static-web-server/src/handler.rs:276:25
    |
276 |                     let mut resp = None;
    |                         ----^^^^
    |                         |
    |                         help: remove this `mut`
```

This should be solved by shadowing the `resp` variable.

palant:
> This should be solved by shadowing the `resp` variable.

Yes, I’ve already noticed that issue myself and addressed it in #368.

src/headers_ext/accept_encoding.rs
palant: Done.

src/compression.rs
palant: Done.

joseluisq (approved):
Thanks!

PR #365, opened 2024-04-25
Use sailfish templates for directory listings

palant:
## Description
This uses sailfish crate to generate JSON and HTML directory listings instead of piecing things together manually. The advantage of sailfish is that it translates templates into Rust code at compile time which makes it fast. Local performance measurements indicate no change for HTML and a slight improvement (up to 5%) for JSON. The added benefit is automatic escaping in HTML. Downside is that `rm_whitespace` option only removes space characters and not newlines, so there are some of those in the listings now.

One thing to consider would be enabling the `json` feature for sailfish and using the corresponding filter instead of `json_quote_str()`. This functionality depends on serde_json which is already in the dependency tree.

## How Has This Been Tested?
Integration tests pass after being adjusted for the added newlines. Functional testing on Linux is also fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1093,6 +1093,8 @@ dependencies = [
  "itoap",
  "ryu",
  "sailfish-macros",
+ "serde",
+ "serde_json",
  "version_check",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,6 +387,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


palant:
> Local performance measurements indicate no change for HTML

Github Action here shows the same thing. Good to have this confirmed, I actually expected a slight performance reduction.

palant:
> One thing to consider would be enabling the `json` feature for sailfish and using the corresponding filter instead of `json_quote_str()`. This functionality depends on serde_json which is already in the dependency tree.

I’ve added this change as well now. Local performance measurement indicates another slight improvement, so no reason to keep the custom code around.

palant:
I realized too late that sailfish is essentially abandoned. So this is superseded by #367 now, maud is under active development and provides better syntax anyway.

PR #368, opened 2024-04-25
Fix errors due to "unused code" when features are disabled

palant:
## Description
Unfortunately, my previous changes caused some compile errors when compiling with `--no-default-features`. That’s mostly due to unused code, one error is because of a variable being unnecessarily mutable.

## Related Issue
#340, #342

## How Has This Been Tested?
Tests pass.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -6,6 +6,14 @@
 //! Request handler module intended to manage incoming HTTP requests.
 //!
 
+#[cfg(any(
+    feature = "compression",
+    feature = "compression-gzip",
+    feature = "compression-brotli",
+    feature = "compression-zstd",
+    feature = "compression-deflate"
+))]
+use headers::HeaderValue;
 use hyper::{Body, Request, Response, StatusCode};
 use std::{future::Future, net::IpAddr, net::SocketAddr, path::PathBuf, sync::Arc};
 
@@ -16,7 +24,7 @@ use std::{future::Future, net::IpAddr, net::SocketAddr, path::PathBuf, sync::Arc
     feature = "compression-zstd",
     feature = "compression-deflate"
 ))]
-use {crate::compression, headers::HeaderValue};
+use crate::compression;
 
 #[cfg(feature = "basic-auth")]
 use crate::basic_auth;
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -289,7 +289,6 @@ impl RequestHandler {
                         &self.opts.page50x,
                     )?;
 
-
                     // Check for a fallback response
                     #[cfg(feature = "fallback-page")]
                     let resp = if req.method().is_get()
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -280,7 +280,15 @@ impl RequestHandler {
             {
                 Ok(result) => (result.resp, result.is_precompressed, Some(result.file_path)),
                 Err(status) => {
-                    let resp = None;
+                    // Produce an error response by default
+                    let resp = error_page::error_response(
+                        req.uri(),
+                        req.method(),
+                        &status,
+                        &self.opts.page404,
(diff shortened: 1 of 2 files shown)


src/handler.rs
```diff
@@ -17,7 +16,7 @@ use std::{future::Future, net::IpAddr, net::SocketAddr, path::PathBuf, sync::Arc
     feature = "compression-zstd",
     feature = "compression-deflate"
 ))]
-use crate::compression;
+use {crate::compression, headers::HeaderValue};
```
joseluisq: Please restore this change and line 9 above too.  It is because I prefer to distinguish between local and external crates.

Btw I will see if I can enforce this via some lint rule or so.

src/handler.rs
```diff
@@ -273,17 +272,20 @@ impl RequestHandler {
...
                         && !self.opts.page_fallback.is_empty()
                     {
-                        resp = Some(fallback_page::fallback_response(&self.opts.page_fallback));
-                    }
+                        Some(fallback_page::fallback_response(&self.opts.page_fallback))
+                    } else {
+                        resp
+                    };
 
+                    #[allow(clippy::unnecessary_literal_unwrap)]
```
joseluisq: I wonder if you even need this. Did you get lint errors without it? I bet that it compiles just fine.

joseluisq (changes requested):

src/handler.rs
palant: Yes, clipply complains when run with `--no-default-features`. It still compiles of course, but I just don’t want the noise.

src/handler.rs
palant: Done.

src/handler.rs
joseluisq: Strange because for me `cargo build --no-default-features` builds with no clippy errors.

joseluisq: Btw `cargo build --no-default-features` could be a good addition to the CI `test` matrix in [`devel.yaml`](https://github.com/static-web-server/static-web-server/blob/master/.github/workflows/devel.yml#L247).
Since I want to improve CI anyway, I will take care of this.

palant: > Strange because for me `cargo build --no-default-features` builds with no clippy errors.

Maybe because I am on nightly.

joseluisq: Tested on nightly but I do not see the error. Can we remove this for now? If fails then we can fix it later. Because for stable compiles just fine.

joseluisq (changes requested):

palant:
> Because for stable compiles just fine.

As I said, this isn’t about compiling but about clippy. Using latest stable (1.77.2) I get when running `cargo clippy --no-default-features`:

```
error: used `unwrap_or()` on `None` value
   --> src/handler.rs:296:32
    |
296 |                       let resp = resp.unwrap_or(
    |  ________________________________^
297 | |                         // Otherwise return an error response
298 | |                         error_page::error_response(
299 | |                             req.uri(),
...   |
304 | |                         )?,
305 | |                     );
    | |_____________________^
    |
help: remove the `None` and `unwrap_or()`
   --> src/handler.rs:283:32
    |
283 |                     let resp = None;
    |                                ^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_literal_unwrap
note: the lint level is defined here
   --> src/lib.rs:99:9
    |
99  | #![deny(warnings)]
    |         ^^^^^^^^
    = note: `#[deny(clippy::unnecessary_literal_unwrap)]` implied by `#[deny(warnings)]`
```

palant:
Whatever, I’ve rearranged this code so that we don’t have any warnings here. The optimizer can figure it out.

joseluisq (approved):
Thanks!

Issue #362, opened 2024-04-23
Static compression should not be dependent on the compression features


palant:
### Search for duplicate feature request

- [X] I already searched, and this feature request or improvement is not a duplicate.

### Feature scope

Improve existing functionality

### Feature request related to a problem

As things are now, static compression chooses which compression types to support based on compression features. This makes little sense given that compression features activate dependencies that static compression doesn’t use. It also means that static compression is occasionally unused when it really should be: `Accept-Encoding: br, gzip` results in SWS only looking for a pre-compressed Brotli file. If it doesn’t exist, the existing gzip file is ignored.

### Describe the solution you'd like

Static compression should either be available unconditionally (only behind the configuration setting) or behind a `compression-static` feature. As it doesn’t introduce any dependencies, the former should be preferable.

Rather than using `compression::get_preferred_encoding()`, static compression should call `AcceptEncoding::sorted_encodings()` directly. For each encoding returned, it should check whether a file with the corresponding extension exists – use it if present, check next encoding if not.

Optionally, a new configuration option could be introduced to restrict the encodings to be considered, encodings not on this list would be skipped over. This is to avoid hitting the file system unnecessarily (e.g. `Accept-Encoding: br, deflate, zstd, gzip`, yet the directory is known to only contain gzip files).

### Describe alternatives you've considered

-

### Build target

All targets

joseluisq:
It makes sense. `compression-static` is dictated by the `compression` feature right now which should be not the case since it mostly works with the `AcceptEncoding` value.
And I agree that a new configuration option could be necessary to limit encodings for efficiency.

msuarezd:
Hola @joseluisq . As I already looked into the compression code, would you mind if I take this?

joseluisq:
Hey @msuarezd. Sure, this one was a good suggestion.

PR #369, opened 2024-04-27
Move all compression-related code into compression and compression_static modules

palant:
## Description
This establishes an analogous internal API for compression as for other modules. It is slightly different for `compression::post_process()` – instead of taking a mutable response reference, it takes a response and returns a new instance. It probably makes sense to adjust other `post_process()` functions to do the same later, for consistency.

There is a merge conflict with #368 here unfortunately. Also, removing `StaticFileResponse::is_precompressed` field is a breaking change. Let me know if you’d prefer me to keep it despite being unused.

## Related Issue
One more step towards #342

## How Has This Been Tested?
Tests pass, functional testing on Linux looks good.

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -22,7 +22,7 @@ use futures_util::Stream;
 use headers::{ContentType, HeaderMap, HeaderMapExt, HeaderValue};
 use hyper::{
     header::{CONTENT_ENCODING, CONTENT_LENGTH},
-    Body, Method, Response,
+    Body, Method, Request, Response, StatusCode,
 };
 use lazy_static::lazy_static;
 use mime_guess::{mime, Mime};
@@ -33,9 +33,11 @@ use std::task::{Context, Poll};
 use tokio_util::io::{ReaderStream, StreamReader};
 
 use crate::{
+    error_page,
+    handler::RequestHandlerOpts,
     headers_ext::{AcceptEncoding, ContentCoding},
     http_ext::MethodExt,
-    Result,
+    Error, Result,
 };
 
 lazy_static! {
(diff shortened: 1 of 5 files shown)


src/static_files.rs
```diff
@@ -83,8 +83,6 @@ pub struct HandleOpts<'a> {
 pub struct StaticFileResponse {
     /// Inner HTTP response.
     pub resp: Response<Body>,
-    /// If the inner HTTP response is already pre-compressed.
-    pub is_precompressed: bool,
```
joseluisq, line 87: > There is a merge conflict with https://github.com/static-web-server/static-web-server/pull/368 here unfortunately. Also, removing StaticFileResponse::is_precompressed field is a breaking change. Let me know if you’d prefer me to keep it despite being unused.

Yes, this is a breaking change. But it's also fine to supersede it by checking the `Content-Encoding` instead. However, this has to be mentioned in the next release.

joseluisq, line 87: > There is a merge conflict with https://github.com/static-web-server/static-web-server/pull/368 here unfortunately. Also, removing StaticFileResponse::is_precompressed field is a breaking change. Let me know if you’d prefer me to keep it despite being unused.

Yes, this is a breaking change. But it's also fine to supersede it by checking the `Content-Encoding` instead. However, this has to be mentioned in the next release.

src/compression.rs
```diff
@@ -64,6 +66,63 @@ const AVAILABLE_ENCODINGS: &[ContentCoding] = &[
...
+    let is_precompressed = resp.headers().get(CONTENT_ENCODING).is_some();
+    if is_precompressed {
+        return Ok(resp);
+    }
+
+    // Compression content encoding varies so use a `Vary` header
+    resp.headers_mut().append(
+        hyper::header::VARY,
+        HeaderValue::from_name(hyper::header::ACCEPT_ENCODING),
+    );
```
joseluisq, line 108: I just noted that we are applying the header twice (also `src/compression_static.rs`) in case `--compression-static` is enabled, which is not by default but maybe we can add this header to some post-processing function to do it once if possible (it was the intention of the boolean flag after all). 

I also assume that this should be more clear once we tackle https://github.com/static-web-server/static-web-server/issues/362 later.

src/compression_static.rs
```diff
@@ -25,6 +29,29 @@ pub struct CompressedFileVariant<'a> {
...
+/// Initializes static compression.
+pub fn init(enabled: bool, handler_opts: &mut RequestHandlerOpts) {
+    handler_opts.compression_static = enabled;
+    server_info!("compression static: enabled={enabled}");
+}
+
+/// Post-processing to add Vary header if necessary.
+pub(crate) fn post_process(
+    opts: &RequestHandlerOpts,
+    _req: &Request<Body>,
```
joseluisq, line 41: How about this unused `_req `? I also saw it in `src/security_headers.rs`. Do you have something in mind for it?

joseluisq (changes requested):
Good overall, I left minor comments to check and the conflicts to address.

src/compression_static.rs
palant: I merely want the call signatures to be consistent. Right now this is irrelevant of course. But the goal is moving towards calling handlers dynamically – be it via middleware or via some trait.

src/compression.rs
palant: Yes, we are applying it twice – but only in the comparably rare scenario where both static and dynamic compression are enabled. I don’t see an easy way to deal with that without intermingling `compression` and `compression_static` modules again when the goal here was actually making these modules self-contained. I only see a way that will introduce significant complexity.

There is no user-visible effect, the second call simply overwrites the first. And the performance impact should be negligible as well. So IMHO not worth addressing.

joseluisq (approved):
Thanks!

PR #371, opened 2024-04-28
Don't percent-encode unreserved characters for directory listing links

joseluisq:
## Description

Previously we were percent-encoding the whole file and directory entry links when listing a directory regardless if those characters can be allowed to be in a URI "as is".

This PR prevents percent-encoding the following characters (a.k.a unreserved marks): `_-.~`

https://www.ietf.org/rfc/rfc3986.txt
## Related Issue

https://github.com/static-web-server/static-web-server/issues/332#issuecomment-2075858645

## Motivation and Context

Improve directory listing

## How Has This Been Tested?

Functional testing and integration tests adjusted accordingly.

### DIFF
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -13,7 +13,7 @@ use headers::{ContentLength, ContentType, HeaderMapExt};
 use humansize::FormatSize;
 use hyper::{Body, Method, Response, StatusCode};
 use mime_guess::mime;
-use percent_encoding::{percent_decode_str, utf8_percent_encode, NON_ALPHANUMERIC};
+use percent_encoding::{percent_decode_str, utf8_percent_encode, AsciiSet, NON_ALPHANUMERIC};
 use serde::{Serialize, Serializer};
 use std::future::Future;
 use std::io;
@@ -21,6 +21,21 @@ use std::path::Path;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::{http_ext::MethodExt, Context, Result};
+
+/// Non-alphanumeric characters to be percent-encoded
+/// excluding the "unreserved marks" because allowed in a URI.
+/// See https://www.ietf.org/rfc/rfc2396.txt
+const PERCENT_ENCODE_SET: &AsciiSet = &NON_ALPHANUMERIC
+    .remove(b'_')
+    .remove(b'-')
+    .remove(b'!')
+    .remove(b'.')
+    .remove(b'(')
+    .remove(b')')
+    .remove(b'*')
+    .remove(b'/')
+    .remove(b'\'')
+    .remove(b'~');
 
 #[derive(Debug, Serialize, Deserialize, Clone, ValueEnum)]
 #[serde(rename_all = "lowercase")]
@@ -75,9 +90,7 @@ pub fn auto_index(
                 opts.dir_listing_order,
                 opts.dir_listing_format,
                 opts.ignore_hidden_files,
-            )
-            .await
-            {
+            ) {
                 Ok(resp) => Ok(resp),
                 Err(err) => {
                     tracing::error!("error after try to read directory entries: {:?}", err);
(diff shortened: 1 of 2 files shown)


joseluisq:
Performance wise I do not see any significant changes.

src/directory_listing.rs
```diff
@@ -183,17 +194,11 @@ async fn read_dir_entries(
...
-            Err(err) => {
-                tracing::error!(
-                    "unable to resolve name for file or directory entry (skipped): {:?}",
-                    err
-                );
+        let entry_name = dir_entry.file_name();
+        let name = match entry_name.to_str() {
+            Some(s) => s.to_owned(),
+            None => {
+                tracing::error!("unable to resolve name for file or directory entry (skipped)");
```
palant: That message isn’t actually correct. We have a non-Unicode file name here, so we cannot convert it to a string.

Skipping these file names is better than failing altogether but still not great. The proper solution would be sticking to `OsString` here. We can use `OsString::to_string_lossy()` when the name needs to be displayed, it will give us the best possible string representation of the name.

We should also use `percent_encode::percent_encode` on `entry_name.as_encoded_bytes()`. This will always work correctly, but I suspect that the file won’t open due to similar issues in `static_file.rs`.

Feel free to implement this, or just leave things as they are and I’ll implement this later.

src/directory_listing.rs
```diff
@@ -241,7 +246,7 @@ async fn read_dir_entries(
                 }
             };
             if symlink_meta.is_dir() {
-                name_encoded.push('/');
+                uri.push('/');
```
palant: There is an additional issue here: we append that slash here but later we don’t consider these symlinks directories (because we only look at `meta.is_dir()` to determine type). Maybe instead of pushing this slash we could set `r#type` here:

```rust
let r#type = if meta.is_dir() {
  dirs_count += 1;
  FileType::DIRECTORY
} else if meta.is_file() {
  …
};
```

Then we can percent-encode the name, append it to the base URI and only then add the slash for directories.

Additional benefit: you won’t have to assume that file names cannot contain slashes. `file\name.txt` is actually a valid file name on Linux, and it needs to be encoded as `file%5Cname.txt` in the URL. Keeping a literal backslash will break the URL.

src/directory_listing.rs
```diff
@@ -303,14 +309,15 @@ async fn read_dir_entries(
             FileType::File
         };
 
-        file_entries.push(FileEntry {
+        let uri = utf8_percent_encode(&uri, PERCENT_ENCODE_SET).to_string();
```
palant: This is a bug, you are encoding `base_dir` twice. Please try listing a directory with non-alphanumeric characters in the name.

src/directory_listing.rs
```diff
@@ -22,6 +22,21 @@ use std::time::{SystemTime, UNIX_EPOCH};
...
+    .remove(b'_')
+    .remove(b'-')
+    .remove(b'!')
+    .remove(b'.')
+    .remove(b'(')
+    .remove(b')')
+    .remove(b'*')
+    .remove(b'/')
+    .remove(b'\'')
+    .remove(b'~');
```
palant, line 32: This RFC is obsolete, please refer to https://www.ietf.org/rfc/rfc3986.txt. This means a shorter list of unreserved characters. Also, neither forward nor backward slash belong here (see other comment on how this can be implemented without allowing slashes).

palant (changes requested):
This is going into the right direction, just not far enough yet.

src/directory_listing.rs
joseluisq: Yes,  I restored partially the changes.
The `read_dir_entries` in general has to be refactored to handle non-Unicode file names properly.

Please feel free to go ahead with this.

palant (approved):
LGTM. I can take care of the rest.

src/directory_listing.rs
joseluisq: Yeah, I was too lazy looks like. So reverted that for now. This can be better simplified later.

joseluisq:
@palant thanks for the feedback. Yes, I was too fast when looking at and missing some details.
Definitely, we can do it better!

PR #372, opened 2024-04-28
Move all fallback page logic into the corresponding module

palant:
## Description
I’ve made sure that fallback page logic is implemented in a generic way, with a `post_process()` call for the response. While at it, I’ve also unified the signatures of the various `post_process()` calls (well, almost).

## Related Issue
This is another step towards #342.

## How Has This Been Tested?
Tests pass, a few more tests have been added. Functional testing on Linux looks fine.

### DIFF
--- a/src/fallback_page.rs
+++ b/src/fallback_page.rs
@@ -7,8 +7,40 @@
 //!
 
 use headers::{AcceptRanges, ContentLength, ContentType, HeaderMapExt};
-use hyper::{Body, Response, StatusCode};
+use hyper::{Body, Request, Response, StatusCode};
 use mime_guess::mime;
+use std::path::Path;
+
+use crate::{handler::RequestHandlerOpts, helpers, http_ext::MethodExt, Error};
+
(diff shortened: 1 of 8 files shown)


src/fallback_page.rs
```diff
@@ -7,8 +7,40 @@
...
 
 use headers::{AcceptRanges, ContentLength, ContentType, HeaderMapExt};
-use hyper::{Body, Response, StatusCode};
+use hyper::{Body, Request, Response, StatusCode};
 use mime_guess::mime;
+use std::path::Path;
+
+use crate::{handler::RequestHandlerOpts, helpers, http_ext::MethodExt, Error};
+
+/// Initializes fallpack page processing
```
joseluisq:
```suggestion
/// Initializes fallback page processing
```

joseluisq (changes requested):
Looks fine to me, just a minor typo.

joseluisq (approved):
Thanks!

PR #373, opened 2024-04-28
Move directory listing initialization into the corresponding module

palant:
## Description
Configuration of the directory listings is now performed in `directory_listing.rs`.

## Related Issue
This should be the last change for #342.

## How Has This Been Tested?
Functional testing on Linux looks ok.


joseluisq (dismissed):
Thanks!

joseluisq (approved):

PR #374, opened 2024-04-28
Make sure to deal with malformed UTF-8 in file names

palant:
## Description
This streamlines the logic in directory listings as discussed. It also makes sure to operate with file names as `OsString`, allowing for invalid Unicode in file names. The (potentially lossy) conversion to string only happens when the name needs to be displayed. Quite a few unnecessary error conditions got removed as a side-effect.

Something to consider would be enabling serialization of the `uri` field in JSON. Since the `name` field might have lost information, there is currently no reliable way to deduce the file’s URL from the JSON data.

## Related Issue
#371

## How Has This Been Tested?
On the Linux command line I did:

```bash
echo test > `echo -e "file\x80.txt"`
mkdir `echo -e "dir\x80.txt"`
echo test > `echo -e "dir\x80/test.txt"`
```

I verified that both the directory and the file show up in the listing (with the Unicode placeholder showing up for `\x80`) and both can be opened. The file `text.txt` in the directory can be opened as well.

As to Windows, I’m not sure that creating a file with a non-Unicode name is even possible. I did verify that the fallback employed there works on Linux (for properly encoded file names) but I didn’t perform any testing on Windows.

### DIFF
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -13,8 +13,9 @@ use headers::{ContentLength, ContentType, HeaderMapExt};
 use humansize::FormatSize;
 use hyper::{Body, Method, Response, StatusCode};
 use mime_guess::mime;
-use percent_encoding::{percent_decode_str, utf8_percent_encode, AsciiSet, NON_ALPHANUMERIC};
+use percent_encoding::{percent_decode_str, percent_encode, AsciiSet, NON_ALPHANUMERIC};
 use serde::{Serialize, Serializer};
+use std::ffi::{OsStr, OsString};
 use std::future::Future;
 use std::io;
 use std::path::Path;
@@ -137,16 +138,15 @@ enum FileType {
 /// Defines a file entry and its properties.
 #[derive(Serialize)]
 struct FileEntry {
-    name: String,
-    #[serde(skip_serializing)]
-    name_encoded: String,
+    #[serde(serialize_with = "serialize_name")]
+    name: OsString,
     #[serde(serialize_with = "serialize_mtime")]
     mtime: Option<DateTime<Local>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     size: Option<u64>,
     r#type: FileType,
     #[serde(skip_serializing)]
-    uri: Option<String>,
+    uri: String,
 }
 
 impl FileEntry {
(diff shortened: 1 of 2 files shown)


palant:
This seems to have improved the performance of large directory slightly, around 2%.

palant:
Another observation: with this change symlinks show up with the size of the linked file which is correct IMHO. Previously it was the size of the symlink which makes little sense given that the file is otherwise not indicated as a symlink.

The MIME type is being derived from the file name of the symlink however, not the linked file. I wonder whether that is the right thing to do here.

src/directory_listing.rs
```diff
@@ -258,59 +241,43 @@ fn read_dir_entries(
...
                 }
             };
             if symlink_meta.is_dir() {
-                name_encoded.push('/');
                 dirs_count += 1;
+                (FileType::Directory, None)
             } else {
-                size = Some(meta.len());
                 files_count += 1;
+                (FileType::File, Some(symlink_meta.len()))
```
joseluisq, line 254: > Another observation: with this change symlinks show up with the size of the linked file which is correct IMHO. Previously it was the size of the symlink which makes little sense given that the file is otherwise not indicated as a symlink.

Yeah, good catch!

src/directory_listing.rs
```diff
@@ -258,59 +241,43 @@ fn read_dir_entries(
...
-
-            let mut base_str = String::new();
-            if !base_dir.starts_with("/") {
-                let base_dir = base_dir.to_str().unwrap_or_default();
-                if !base_dir.is_empty() {
-                    base_str.push_str(base_dir);
-                }
-                base_str.push('/');
-            }
+        let mut uri = if !base_path.ends_with('/') && !base_path.is_empty() {
```
joseluisq, line 269: Maybe we can express the conditional like this below just for clearness.

```suggestion
        let mut uri = if !base_path.is_empty() && !base_path.ends_with('/') {
```

src/file_path.rs
```diff
@@ -30,20 +27,29 @@ impl PathExt for Path {
...
+
+    OsStr::from_bytes(bytes).into()
+}
+
+#[cfg(windows)]
+fn path_from_bytes(bytes: &[u8]) -> PathBuf {
+    // This should really be OsStr::from_encoded_bytes_unchecked() but it’s
+    // unsafe. With this fallback non-Unicode file names will result in 404.
+    String::from_utf8_lossy(bytes).into_owned().into()
+}
```
joseluisq, line 43: > As to Windows, I’m not sure that creating a file with a non-Unicode name is even possible. I did verify that the fallback employed there works on Linux (for properly encoded file names) but I didn’t perform any testing on Windows.

Comment: I also don't know how to create such a file but I'm fine with the safe approach of converting the invalid characters anyway.
Btw I just did a general functional testing in Windows.

joseluisq (changes requested):
Just minor comments to check.

joseluisq:
> The MIME type is being derived from the file name of the symlink however, not the linked file. I wonder whether that is the right thing to do here.

Good question, yeah the symlink name could differ from the actual linked file but I do not if we should follow the symlink to that point.
Maybe we could come up with some idea of how other web servers handle this.
But, honestly symlink following should have been an opt-in feature and disabled by default.

joseluisq (approved):
No need to delay the PR.
Thanks!

src/directory_listing.rs
palant: I don’t mind either way. My logic was that checking for the trailing slash was the main check here and `!is_empty()` merely a safeguard that quite frankly should never even happen.

Do you want me to create another PR for this?

joseluisq: All fine!

PR #375, opened 2024-04-28
Do not mention request body type unnecessarily #340

palant:
## Description
This uses generics to avoid mentioning the `Body` type for requests since we don’t actually care about it. This will cut down on the compile errors to be addressed for #340.

Note that I revered the initially implemented changes to `handler.rs` and `service.rs`. The former is being used by integration tests, not specifying the body type there requires specifying it everywhere in the tests – not worth it.

## Related Issue
#340

## How Has This Been Tested?
Automated tests pass.


joseluisq (approved):
> Note that I revered the initially implemented changes to handler.rs and service.rs. The former is being used by integration tests, not specifying the body type there requires specifying it everywhere in the tests – not worth it.

Ok, Let's tackle them later.

Issue #370, opened 2024-04-27
Directory listing: File sizes wrongly displayed in decimal format


palant:
### Search for duplicate issues

- [X] I already searched, and this issue is not a duplicate.

### Issue scope

Build target (binary)

### Describe the bug

SWS uses the decimal format for displaying file sizes: 1 KB = 1000 bytes, 1 MB = 1000 MB. This makes no sense.

### How to reproduce it

1. Put a file with 1230 bytes into a directory.
2. Check the directory listing.

The file size will be displayed as “1.23 KB”

### Expected behavior

The file size should be displayed as “1.20 KiB” using the binary format (1 KiB = 1024 bytes). Note that the trailing zero is important to make numbers easier to read, humansize will currently skip it.

### Complementary information

I checked the other web servers. Apache and lighttpd use the binary format, the former not displaying any decimal places. nginx will simply display the raw number.

### Build target

Built from source (specify below)

### Environment and specs

-

### Additional context

I wonder whether the dependency on humansize is worth it. That’s quite a bit of complexity for something that can be done in a few lines of code, and it shows up in the performance graph quite prominently (not as prominently as chrono though).

joseluisq:
Good catch! The `humanize` crate could be even not needed.

miroim:
Hi, can I have a try?

joseluisq:
> Hi, can I have a try?

Sure, unless @palant is working on this?

palant:
Was just about to start but – sure, feel free to.

joseluisq:
@miroim you probably want to also join Discord to coordinate development.

joseluisq:
Closing as resolved by #376

PR #376, opened 2024-04-28
Fix the display format for file sizes

miroim:
## Description

Replace `humansize` with a function `format_file_size` in `src/directory_listing.rs`
## Related Issue

https://github.com/static-web-server/static-web-server/issues/370

## How Has This Been Tested?

Add some test cases in `src/directory_listing.rs`
Compared and tested with local files

### DIFF
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -3,7 +3,7 @@
 // See https://static-web-server.net/ for more information
 // Copyright (C) 2019-present Jose Quintana <joseluisq.net>
 
-//! It provides directory listig and auto-index support.
+//! It provides directory listing and auto-index support.
 //!
 
 use chrono::{DateTime, Local, Utc};
@@ -219,7 +219,7 @@ fn read_dir_entries(
             (FileType::File, Some(meta.len()))
         } else if meta.file_type().is_symlink() {
             // NOTE: we resolve the symlink path below to just know if is a directory or not.
-            // Hwever, we are still showing the symlink name but not the resolved name.
+            // However, we are still showing the symlink name but not the resolved name.
 
             let symlink = dir_entry.path();
             let symlink = match symlink.canonicalize() {
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -10,7 +10,6 @@ use chrono::{DateTime, Local, Utc};
 use clap::ValueEnum;
 use futures_util::{future, future::Either};
 use headers::{ContentLength, ContentType, HeaderMapExt};
-use humansize::FormatSize;
 use hyper::{Body, Method, Response, StatusCode};
 use mime_guess::mime;
 use percent_encoding::{percent_decode_str, utf8_percent_encode, AsciiSet, NON_ALPHANUMERIC};
@@ -490,7 +489,7 @@ fn html_auto_index<'a>(
                             td align="right" {
                                 (match entry.size.unwrap_or(0) {
                                     0 => "-".to_owned(),
-                                    size => size.format_size(humansize::DECIMAL)
+                                    size => format_file_size(size),
                                 })
                             }
                         }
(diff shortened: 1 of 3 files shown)


src/directory_listing.rs
```diff
@@ -589,3 +588,70 @@ fn parse_last_modified(modified: SystemTime) -> Result<DateTime<Local>> {
...
+
+/// Formats the file size in bytes to a human-readable string
+fn format_file_size(size: u64) -> String {
+    let units = ["B", "KiB", "MiB", "GiB", "TiB", "PiB"];
+    let mut size_tmp = size;
+
+    for i in 0..units.len() - 1 {
+        if size_tmp < 1024 {
+            // return the size with the current unit
+            return format!("{} {}", size_tmp, units[i]);
```
palant: This case doesn’t need to be inside the loop – you can have bytes as a special case before it.

src/directory_listing.rs
```diff
@@ -589,3 +588,70 @@ fn parse_last_modified(modified: SystemTime) -> Result<DateTime<Local>> {
         )),
     }
 }
+
+/// Formats the file size in bytes to a human-readable string
+fn format_file_size(size: u64) -> String {
+    let units = ["B", "KiB", "MiB", "GiB", "TiB", "PiB"];
```
palant: This should be a constant.

src/directory_listing.rs
```diff
@@ -589,3 +588,70 @@ fn parse_last_modified(modified: SystemTime) -> Result<DateTime<Local>> {
...
+            return format!("{} {}", size_tmp, units[i]);
+        } else if size_tmp < 1024 * 1024 && i < units.len() - 1 {
+            // return the size divided by 1024 with the next unit
+            return format!("{:.2} {}", size_tmp as f64 / 1024.0, units[i + 1]);
+        } 
+        size_tmp >>= 10;
+    }
+
+    // if size is too large, return the largest unit
+    format!("{} {}", size_tmp, units[units.len() - 1])
```
palant: We are going to have at most 4096 PiB in 64 bit, not large enough to strike the decimal points. So it’s probably best if we only iterate over `units[1 . .units.len() - 1]` – `units[0]` is handled as a special case up front and `units[units.len() - 1]` is a fallback unit to be used if the loop finishes without determining a suitable unit. You can really iterate over the values, not the indexes in that setup. And you’ll leave the loop with a value that you can divide by 1024.0 to display it.

src/directory_listing.rs
```diff
@@ -589,3 +588,70 @@ fn parse_last_modified(modified: SystemTime) -> Result<DateTime<Local>> {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::directory_listing::format_file_size;
```
palant: Nit: `use super::format_file_size;`

src/directory_listing.rs
```diff
@@ -589,3 +588,70 @@ fn parse_last_modified(modified: SystemTime) -> Result<DateTime<Local>> {
...
+
+    #[test]
+    fn handle_byte() {
+        let size = 128;
+        assert_eq!("128 B", format_file_size(size))
+    }
+    
+    #[test]
+    fn handle_kibibyte() {
+        let size = 1690;
```
palant: Nit: Maybe test 1024 just to see that trailing zeros aren’t removed?

palant (changes requested):
Looking good, and I do feel sorry about all the nitpicking.

miroim:
> Looking good, and I do feel sorry about all the nitpicking.

It's all part of helping me improve, and I appreciate your attention to detail.

src/directory_listing.rs
```diff
@@ -555,3 +554,74 @@ fn parse_last_modified(modified: SystemTime) -> Result<DateTime<Local>> {
...
 }
+
+/// Formats the file size in bytes to a human-readable string
+fn format_file_size(size: u64) -> String {
+    const UNITS: [&str; 6] = ["B", "KiB", "MiB", "GiB", "TiB", "PiB"];
+    let mut size_tmp = size;
+
+    if size_tmp < 1024 {
+        // return the size with Byte
+        return format!("{} {}", size_tmp, "B");
```
palant: Please use `UNITS[0]` as parameter here.

palant (changes requested):
Looks good, one tiny thing left to improve.

palant (approved):
LGTM

joseluisq (approved):
Thanks!

palant:
@miroim FYI, there was an issue with the code that clippy complained about (you have to run it with `--tests` command line flag). I fix it in #382.

PR #384, opened 2024-05-01
Drop custom conversion between SystemTime and DateTime

palant:
## Description
The chrono crate implements conversion from `SystemTime` to `DateTime`, a custom implementation which does mostly the same thing (minus handling timestamps before the Unix epoch) is unnecessary.

## How Has This Been Tested?
Functional testing on Linux is fine.


joseluisq (approved):
Thanks!

PR #439, opened 2024-05-10
Look for all accepted encodings for precompressed files but keeping the quality ordering

### LINKED ISSUE #438: Look for all accepted encodings for precompressed files

### Feature scope

Browsers like Chrome don't provide a relative quality factor for the Accept-Encoding header, but use a fixed order, that does not necessarily express the preference of the browser, but is more likely to ensure backwards compatibility with older web servers. The static-web-server however only checks whether a precompressed file exists for the most preferred, i.e. first accepted encoding.

Keep the same ordering by quality, but look for precompressed files in all accepted encoding standards.

If no quality order is provided in the header, it might be useful to have the web server choose a quality order, e.g. for precompressed files, based on file size. 

### Build target

All targets

### Additional context

_No response_

kobutri:
## Description

Instead of only looking for a precompressed file matching the preferred encoding, I check for all accepted encodings.

## Related Issue

https://github.com/static-web-server/static-web-server/issues/438

## Motivation and Context

Crome accepts brotli and brotli is generally superior to gzip, but  Chrome always lists gzip first in the `Accept-Encoding` header. 
https://github.com/static-web-server/static-web-server/issues/438

## How Has This Been Tested?

I created a new brotli compressed `index.html.br` and created a test that accepts `deflate, br`. Because of this change the `index.html.br` will be returned, where previously no compressed file would be returned.


joseluisq (approved):
The idea is not bad. In fact, it helps SWS to decide better which file to serve.
FYI I just fixed syntax issues and improved the test a bit.

Thanks!

PR #449, opened 2024-05-22
update installer.sh to support overriding default variables

frankli0324:
## Description

adds two optional environment variables controlling the installation script:

```sh
export SWS_INSTALL_DIR="~/.local/bin"
export SWS_INSTALL_VERSION="2.31.0"
curl --proto '=https' --tlsv1.2 -sSfL https://get.static-web-server.net | sh
```

## Motivation and Context

when installing with get.static-web-server.net, I would like to be able to pin the install version and specify an installation directory (so I can use it in packaging scripts)

### DIFF
--- a/docs/content/download-and-install.md
+++ b/docs/content/download-and-install.md
@@ -27,6 +27,20 @@ curl --proto '=https' --tlsv1.2 -sSfL https://get.static-web-server.net | sh
 
 `static-web-server` should be installed under the `/usr/local/bin` directory.
 
+You can install a specific version of SWS to a custom location by setting environment variables:
+
+```sh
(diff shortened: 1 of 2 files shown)


joseluisq (changes requested):
Nice idea. I wonder if you could also update the installer page with an example of how to use it. https://github.com/static-web-server/static-web-server/blob/master/docs/content/download-and-install.tmpl.md#binary-installer-linuxbsds

BTW can you please also update the PR with the example?

joseluisq (approved):
Thanks!

joseluisq:
BTW since it is a script (dynamic) and already merged. It should be already available to be used. The docs should be published soon.

PR #475, opened 2024-08-31
Support for generating man pages and completions

jcgruenhage:
## Description
This adds a way to generate completions and man pages so that distro packagers can ship them.

Instead of doing it based on an env var, which others are doing in f.ex. https://codeberg.org/openpgp-card/openpgp-card-tools/src/branch/main/src/oct.rs#L34-L56, it would also be possible to add a subcommand or a special command line flag, like other are doing in f.ex. https://github.com/orhun/git-cliff/blob/v2.4.0/git-cliff/src/bin/mangen.rs

## Related Issue
I have not opened an issue about this. The change is fairly trivial though. If I really need to open an issue, I can do that, but I don't really see the purpose of it.

## Motivation and Context
I have written this patch while packaging static-web-server for Chimera Linux, and it was (rightfully) pointed out that this patch should go upstream: https://github.com/chimera-linux/cports/pull/2818#pullrequestreview-2273742092.

## How Has This Been Tested?
A slightly different version of this patch has been tested in the cports PR linked above.

#### Fish
image
#### Nushell
image
### Man page
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -263,6 +263,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "carapace_spec_clap"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9c1b382d0ea2f304c9dba34f746284c7b6d231db48eae53b36d1e6eda1aba402"
+dependencies = [
+ "clap",
+ "clap_complete",
+ "indexmap",
+ "serde",
+ "serde_yaml",
+]
+
+[[package]]
 name = "cargo-platform"
 version = "0.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -342,6 +355,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_allgen"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "38b31337b55b534905114db35586730324e5dab9a01af7cf440db39723caeab4"
+dependencies = [
+ "carapace_spec_clap",
+ "clap",
+ "clap_complete",
+ "clap_complete_fig",
+ "clap_complete_nushell",
+ "clap_mangen",
+ "thiserror",
+]
+
+[[package]]
 name = "clap_builder"
 version = "4.5.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -351,6 +379,35 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+]
+
+[[package]]
+name = "clap_complete"
+version = "4.5.26"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "205d5ef6d485fa47606b98b0ddc4ead26eb850aaa86abfb562a94fb3280ecba0"
+dependencies = [
+ "clap",
+]
+
+[[package]]
+name = "clap_complete_fig"
+version = "4.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d494102c8ff3951810c72baf96910b980fb065ca5d3101243e6a8dc19747c86b"
+dependencies = [
+ "clap",
+ "clap_complete",
+]
+
+[[package]]
+name = "clap_complete_nushell"
+version = "4.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5fe32110e006bccf720f8c9af3fee1ba7db290c724eab61544e1d3295be3a40e"
+dependencies = [
+ "clap",
+ "clap_complete",
 ]
 
 [[package]]
@@ -370,6 +427,16 @@ name = "clap_lex"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1462739cb27611015575c0c11df5df7601141071f07518d56fcc1be504cbec97"
+
+[[package]]
+name = "clap_mangen"
+version = "0.2.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f17415fd4dfbea46e3274fcd8d368284519b358654772afb700dc2e8d2b24eeb"
+dependencies = [
+ "clap",
+ "roff",
+]
 
 [[package]]
 name = "colorchoice"
@@ -836,6 +903,7 @@ checksum = "93ead53efc7ea8ed3cfb0c79fc8023fbb782a5432b52830b6518941cebe6505c"
 dependencies = [
  "equivalent",
  "hashbrown",
+ "serde",
 ]
 
 [[package]]
@@ -1287,6 +1355,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "roff"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "88f8660c1ff60292143c98d08fc6e2f654d722db50410e3f3797d40baaf9d8f3"
+
+[[package]]
 name = "rustc-demangle"
 version = "0.1.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


src/bin/server.rs
```diff
@@ -21,6 +21,17 @@
         return static_web_server::settings::cli_output::display_version();
     }
 
+    if let Ok(path) = std::env::var("SWS_GENERATE") {
+        let path = std::path::PathBuf::from(path);
+        let mut completions_path = path.clone();
+        completions_path.push("completions");
+        clap_allgen::render_shell_completions::<cli::General>(completions_path)?;
```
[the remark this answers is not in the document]
jcgruenhage: must've missed this bit during modifying the patch for upstream submission.

I can't build it on Chimera Linux locally without patching out jemalloc, which is why the patch looks different in the package. Should be fixed now.

src/bin/server.rs
```diff
@@ -21,6 +21,19 @@ fn main() -> Result {
         return static_web_server::settings::cli_output::display_version();
     }
 
+    if let Ok(path) = std::env::var("SWS_GENERATE") {
```
joseluisq: I would like to see this functionality behind a new subcommand so it does not get mixed with the server options.

An idea could be:

- `static-web-server generate --completions <OUT_DIR>`
- `static-web-server generate --man-page <OUT_DIR>`
- `static-web-server generate  <OUT_DIR>` (generate both)

What do you think?

src/bin/server.rs
```diff
@@ -21,6 +21,19 @@ fn main() -> Result {
...
+        let path = std::path::PathBuf::from(path);
+        let mut completions_path = path.clone();
+        completions_path.push("completions");
+        clap_allgen::render_shell_completions::<static_web_server::settings::cli::General>(
+            completions_path,
+        )?;
+        let mut man_path = path.clone();
+        man_path.push("man");
+        clap_allgen::render_manpages::<static_web_server::settings::cli::General>(man_path)?;
+        return Ok(());
```
joseluisq: We should log every operation with some `tracing::info` to hint users.

src/bin/server.rs
joseluisq: Dumb question, is this feature intended for Linux/Unix(es) only? How about Windows? If it is only for *nixes then we could provide it under a compile-time `cfg` macro.

joseluisq (changes requested):
I like the idea. I left a few suggestions and a question to get your input about.

src/bin/server.rs
jcgruenhage: Sounds good to me!

jcgruenhage: I'm not sure whether Windows has man pages, but at least the completions include completions for powershell, so at least that bit is not specific to the *nixes

src/bin/server.rs
jcgruenhage: sounds good to me, will do.

src/bin/server.rs
joseluisq: Good, then let's make man pages available only for *nixes and completions for all OSes.

jcgruenhage: not supporting man page generation on windows makes this quite a bit more complicated I think. How important is that to you?

joseluisq: My point was if man pages are not supported in Windows then we can disable it. Otherwise, if supported then let's leave it like you said.

joseluisq: In the end, a man page is just a text file right? Does not matter if windows is able to read it or not. 😅

joseluisq: Just let me know if you need a review or help.

jcgruenhage: Yeah, it's just generating a text file that might not be readable by any installed software. Some people do install groff (the format behind man pages) readers on Windows as well, so it really doesn't matter.

This is the last thing where I wasn't sure, the rest should be good to go I think.

I haven't tested compilation on Windows yet, but from my reading of the code it *should* work. If you want I can look into getting a VM to try it out.

If you're happy with man pages being generated for windows as well, then this is ready for review now.

src/bin/server.rs
```diff
@@ -21,25 +24,43 @@ fn main() -> Result {
...
+                    clap_allgen::render_shell_completions::<General>(&comp_dir)?;
+                    tracing::info!("Wrote completions to {}", comp_dir.to_string_lossy());
                 }
-                Commands::Uninstall {} => {
-                    return winservice::uninstall_service();
+                if man_pages || (!completions && !man_pages) {
+                    let mut man_dir = out_dir.clone();
+                    man_dir.push("man");
+                    clap_allgen::render_manpages::<General>(&man_dir)?;
+                    tracing::info!("Wrote man pagess to {}", man_dir.to_string_lossy());
```
joseluisq:
```suggestion
                    tracing::info!("Wrote man pages to {}", man_dir.to_string_lossy());
```

src/bin/server.rs
```diff
@@ -21,25 +24,43 @@ fn main() -> Result {
...
+            } => {
+                if completions || (!completions && !man_pages) {
+                    let mut comp_dir = out_dir.clone();
+                    comp_dir.push("completions");
+                    clap_allgen::render_shell_completions::<General>(&comp_dir)?;
+                    tracing::info!("Wrote completions to {}", comp_dir.to_string_lossy());
                 }
-                Commands::Uninstall {} => {
-                    return winservice::uninstall_service();
+                if man_pages || (!completions && !man_pages) {
```
joseluisq:
```suggestion
                if man_pages || !completions {
```

src/bin/server.rs
```diff
@@ -21,25 +24,43 @@ fn main() -> Result {
...
+            #[cfg(windows)]
+            Commands::Uninstall {} => {
+                return winservice::uninstall_service();
+            }
+            Commands::Generate {
+                completions,
+                man_pages,
+                out_dir,
+            } => {
+                if completions || (!completions && !man_pages) {
```
joseluisq:
```suggestion
                if completions || !man_pages {
```

Cargo.lock
joseluisq, line 1: There is a conflict to address here.

joseluisq (changes requested):
It looks fine to me I just left a few minor improvements to do.

Just the last thing to do is create a feature page called `man-pages-completions.md` in [`/docs/content/features`](https://github.com/static-web-server/static-web-server/tree/master/docs/content/features) dir.
Then add a _feature_ entry in the docs config file [`/docs/mkdocs.yml`](https://github.com/static-web-server/static-web-server/blob/master/docs/mkdocs.yml#L168) below the last entry.

Please mention in the docs how to use the new subcommand and what can be generated for the three cases.

After that, we should be able to merge.

src/bin/server.rs
joseluisq: I didn't test on Windows yet but it should work.

jcgruenhage:
@joseluisq I think I've addressed all your feedback. Can you check again whether you're happy with this as it is?

src/bin/server.rs
```diff
@@ -21,25 +24,43 @@ fn main() -> Result {
         return static_web_server::settings::cli_output::display_version();
     }
 
-    #[cfg(windows)]
-    {
-        use static_web_server::settings::Commands;
-        use static_web_server::winservice;
```
joseluisq, line 27: FYI I just fixed this since it broke the Windows build.

joseluisq:
I also fixed the log info.

```sh
$ .\static-web-server.exe generate .
2024-09-08T19:01:16.740824Z  INFO static_web_server::info: wrote completions to .\completions
2024-09-08T19:01:16.757544Z  INFO static_web_server::info: wrote man pages to .\man
```

```sh
/static-web-server generate --completions .`
2024-09-08T19:03:46.665862Z  INFO static_web_server::info: wrote completions to ./completions
```

joseluisq (approved):
Thanks!

joseluisq:
> I have written this patch while packaging static-web-server for Chimera Linux, and it was (rightfully) pointed out that this patch should go upstream: [chimera-linux/cports#2818 (review)](https://github.com/chimera-linux/cports/pull/2818#pullrequestreview-2273742092).

@jcgruenhage FYI if the package gets merged upstream then you probably want to include it in the [SWS Installations](https://static-web-server.net/download-and-install/#installation-methods) page as other distros, just feel free to open a docs PR.

PR #486, opened 2024-09-29
Truenas deprecated

ctag:
Removes now deprecated installation instructions for Truenas/Truecharts.
https://truecharts.org/news/scale-deprecation/


joseluisq:
Thanks for the reminder. It seems like the Helm charts are still being maintained. https://github.com/truecharts/charts/tree/master/charts/stable/static-web-server

> **So, does this mean that TrueCharts is dead?**
> No, TrueCharts is not dead. We have always made Helm Charts and our Charts can be installed on any Helm-compatible OS, not just TrueNAS Scale. We continue to provide these Helm Charts as always and they can already be installed on a compatible OS.
> https://truecharts.org/news/scale-faq/

Maybe we have to make this clear in the docs instead of removing the entry?

ctag:
That's good at least, but helm charts aren't directly usable as Truenas installation material. Possibly just rename that section away from Truenas Scale.

For context, installing apps on Scale through the truecharts catalog is now fully broken. See: https://www.reddit.com/r/truenas/comments/1e1dxvk/truecharts_just_nuked_everything_in_response_to/
So even though the helm charts are maintained, they aren't usable for Scale.

joseluisq:
> For context, installing apps on Scale through the truecharts catalog is now fully broken. See: https://www.reddit.com/r/truenas/comments/1e1dxvk/truecharts_just_nuked_everything_in_response_to/
> So even though the helm charts are maintained, they aren't usable for Scale.

Fair, then let's remove any _TrueNAS Scale_ mention and adjust the section accordingly.

docs/content/features/docker.md
```diff
@@ -189,6 +189,3 @@ spec:
           port: http
 ```
 
-## TrueNAS SCALE
-
-If you use Kubernetes with [TrueNAS SCALE](https://www.truenas.com/truenas-scale/) then check out the [TrueCharts Community Website](https://truecharts.org/charts/stable/static-web-server/) and its [Introduction to SCALE](https://truecharts.org/manual/SCALE/guides/scale-intro) page to deploy an SWS application in your instance.
```
joseluisq, line 194:
```suggestion
### TrueCharts
The [TrueCharts Community](https://truecharts.org) also provides a ready-to-use [Static Web Server Helm-Chart](https://truecharts.org/charts/stable/static-web-server/) that you can easily deploy in your Kubernetes.
```

joseluisq (changes requested):
I just left a minor adjustment to make.
After that, we should merge the PR.

joseluisq (approved):
Thanks!

PR #496, opened 2024-11-06
feat: uds

### LINKED ISSUE #484: Support Unix Domain Socket

### Feature scope

Completely new feature

This is useful when the web-server sits behind a gateway for more efficient communication.

Support Unix Domain Socket

Hyper + Axum + Towerhttp

### Build target

All targets

### Additional context

_No response_

Jeidnx:
## Description

This is a very crude proof of concept for implementing unix domain socket support. Currently only uds works and the path is hardcoded to `/tmp/sws.sock`. I wanted to have some feedback first, before fully implementing this. Here are a few points i would like to have feedback on / discuss:

 - How should the different server types be handled in the configuration? Currently there is a boolean to enable http2, but this wont work with three distinct types. A simple solution would be to replace the `http2` boolean with a `listen_type` (or similar) String. 
 - The `start_server` method is a bit too long imo, does it make sense to have seperate start functions for each type? On the other hand there is already some duplicate code in there, and splitting that into functions would make it harder to reuse code across listeners.

## Related Issue

closes #484
## Motivation and Context

See #484
I mainly want this because remembering paths is simpler than remembering ports, and for the performance.

## How Has This Been Tested?

Tested functionality with curl, works as expected.


joseluisq:
> * How should the different server types be handled in the configuration? Currently there is a boolean to enable http2, but this wont work with three distinct types. A simple solution would be to replace the `http2` boolean with a `listen_type` (or similar) String.

It can be the three types, but let me think a bit about it, maybe we do not need to do this. I will come back to this point soon.

> * The `start_server` method is a bit too long imo, does it make sense to have seperate start functions for each type? On the other hand there is already some duplicate code in there, and splitting that into functions would make it harder to reuse code across listeners.

Yes, that method is convoluted and we are aware of that. About this, there was an attempt to extract functionality and improve the code, for example, https://github.com/static-web-server/static-web-server/pull/377, but unfortunately got stuck.
It makes total sense to start moving out stuff to proper functions/modules and simplifying the code as well.

The question is, do you want to do it before UDS or after?

Jeidnx:
> Server types

Im sure there doesn't need to be a `server_type` key. One option would be to create `http_listener`, `http2_listener` and `socket_listener` options which are mutually exclusive for example. Though im not sure how easy doing mutually exclusive options is with the current configuration file impl. I think having an explicit `server_type` is the easiest to implement and understand from a user perspective. I may have missed something though.

> Refactoring `start_server`

I think it makes sense to do the refactoring together with adding uds support. The uds part isn't that big of a change i think so the changes should still be manageable.
My rough idea is to add the `ServerType` enum to the `Server` struct and turn it into a builder of some sorts. I should probably find some time later this week to really start working on this.

joseluisq:
> Im sure there doesn't need to be a `server_type` key. One option would be to create `http_listener`, `http2_listener` and `socket_listener` options which are mutually exclusive for example. Though im not sure how easy doing mutually exclusive options is with the current configuration file impl. I think having an explicit `server_type` is the easiest to implement and understand from a user perspective. I may have missed something though.

Honestly, It sounds fine to me. I also think this should allow us to decouple `tls` from `http2` and deprecate `--http2-cert/key` in favor of more concise options like `--tls`, `--tls-cert`, `--tls-key` in the future (just an idea).

But keep in mind do not disrupt existing functionality and properly advise users to give them a proper path to migrate if the case.

> I think it makes sense to do the refactoring together with adding uds support. The uds part isn't that big of a change i think so the changes should still be manageable.
> My rough idea is to add the `ServerType` enum to the `Server` struct and turn it into a builder of some sorts. I should probably find some time later this week to really start working on this.

Feel free to go ahead. A prototype of this should be good to see.
Just take into account not to complicate yourself too much in the first phase. Maybe you want to split up the work in recessive PRs before or after UDS.

If you need any help just let me know.

joseluisq:
What about an interface like this below that could convert from `PathBuf::from(s)` or `s.parse::<IpAddr>()?` (just an idea).

```rs
pub enum Interface {
    Address(IpAddr),
    Path(PathBuf),
}
```

joseluisq:
FYI, the Unix Domain Sockets feature is already implemented by https://github.com/static-web-server/static-web-server/pull/681, part of the upcoming v3.

PR #501, opened 2024-11-27
Prevent single wildcards from matching path separators for URL Redirects

mschoettle:
## Description

`globset` by default has `literal_separator(false)` leading to `*` matching the path separator.
This PR ensures that `*` does not match the path separator in redirects.

See: https://docs.rs/globset/0.4.15/globset/#syntax

## Related Issue

See the documentation here: https://github.com/orgs/static-web-server/discussions/498

## Motivation and Context

As per the discussion, the default of not matching the path separator is in line with how it works on Unix.

See also: https://en.wikipedia.org/wiki/Glob_(programming) and https://man7.org/linux/man-pages/man7/glob.7.html#:~:text=A%20'/'%20in%20a%20pathname%20cannot%20be%20matched

## How Has This Been Tested?

Via test cases and manually.

image

### DIFF
--- a/docs/content/features/url-redirects.md
+++ b/docs/content/features/url-redirects.md
@@ -34,6 +34,12 @@ The glob pattern functionality is powered by the [globset](https://docs.rs/globs
 !!! tip "Glob pattern syntax"
     For more details about the Glob pattern syntax check out https://docs.rs/globset/latest/globset/#syntax
 
+!!! warning "Matching of path separator in `*`"
+    Up to version 2.33.1 the wildcard `*` was matching the path separator.
+    For example, `/{*}/{*}/` matched `/assets/images/logo/`.
+
+    In later versions, the default has changed such that `*` does not match the path separator.
+
 ### Destination
 
 The value can be either a local file path that maps to an existing file on the system or an external URL.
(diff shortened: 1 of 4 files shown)


mschoettle:
@joseluisq 

With `literal_seprator(true)` the following test cases don't work as before anymore:

```toml
# Glob groups 5
[[advanced.redirects]]
source = "**/{*}.{jpg,jpeg}"
destination = "http://localhost/new-images/$2.$3"
kind = 302

# Glob groups 6
[[advanced.redirects]]
source = "**/{*}.{ttf,otf,woff}"
destination = "http://localhost/new-fonts/$2.woff"
kind = 302
```

For instance:

```
thread 'tests::redirects_glob_groups_5' panicked at tests/redirects.rs:158:17:
assertion `left == right` failed
  left: "http://localhost/new-images/avatar.jpeg"
 right: "http://localhost/new-images/images/avatar.jpeg"
```

The additional path is now contained in `$1` from `**`. It doesn't seem that it be appropriate to change the assertion of the tests. Maybe with also adding additional paths into the requested URL to test the `**`?

joseluisq:
> The additional path is now contained in `$1` from `**`. It doesn't seem that it be appropriate to change the assertion of the tests. Maybe with also adding additional paths into the requested URL to test the `**`?

You can try that. I wonder if we can take advantage of this PR and add a few more tests tackling other common cases. E.g. double/single `*`, ranges, etc.

--

Additionally, since `literal_seprator(true)` will be a breaking change for the sake of consistency, we will need also to inform users properly via the [Redirects page](https://github.com/static-web-server/static-web-server/blob/master/docs/content/features/url-redirects.md).

Aside, I'm still wondering why _globset_ does not go with `literal_seprator(true)` by default.

mschoettle:
> I wonder if we can take advantage of this PR and add a few more tests tackling other common cases. E.g. double/single *, ranges, etc.

Sure. Can you add more detail what you mean with those examples? It’s not totally clear to me.

joseluisq:
> Sure. Can you add more detail what you mean with those examples? It’s not totally clear to me.

I just meant to add a couple of other unit test cases (not just `/{*}/{*}/`) with group combinations using ranges, for example `/[0-9]/[a-z]/{*}/`,  `/crop-[^0-9]/*.{jpg,jpeg}` or another.

mschoettle:
I fixed the tests.

> Additionally, since literal_seprator(true) will be a breaking change for the sake of consistency, we will need also to inform users properly via the [Redirects page](https://github.com/static-web-server/static-web-server/blob/master/docs/content/features/url-redirects.md).

Added an extra admonition warning about this. Happy for any alternative suggestions on the text etc. Is the changelog automatically generated? Otherwise we would need to add it there too.

> Aside, I'm still wondering why globset does not go with literal_seprator(true) by default.

I wonder the same. So I asked here: https://github.com/BurntSushi/ripgrep/discussions/2940

Regarding the other tests. If you're ok with it I would do this in a separate PR. There are also some warnings that mkdocs reports that would be good to address.

joseluisq:
> Is the changelog automatically generated? Otherwise we would need to add it there too.

The changelog is not automatically generated but we will mention it in the upcoming release for sure.

> Regarding the other tests. If you're ok with it I would do this in a separate PR. There are also some warnings that mkdocs reports that would be good to address.

That makes sense, take your time. The current PR though looks fine to me.

joseluisq (approved):
Thanks!

joseluisq:
I'm also thinking that the URL Rewrites feature should follow the same approach. PR is welcome.

mschoettle:
> I'm also thinking that the URL Rewrites feature should follow the same approach. PR is welcome.

Makes sense. Is this required for the next release?

joseluisq:
> Makes sense. Is this required for the next release?

No. It isn't, but I have plans to release v2.34.0 on the weekend but take your time.
If you would to work on that and want it on the next release let me know.

Btw feel free to join Discord.

PR #503, opened 2024-11-29
Add additional tests for URL Redirect feature

mschoettle:
## Description

Follow-up to #501 to add more redirect tests.

## Related Issue

#501

## Motivation and Context

More tests for various use cases for redirects.

## How Has This Been Tested?

`cargo test`

n/a

### DIFF
--- a/tests/fixtures/toml/redirects.toml
+++ b/tests/fixtures/toml/redirects.toml
@@ -47,6 +47,19 @@ source = "**/{*}.{ttf,otf,woff}"
 destination = "http://localhost/new-fonts/$2.woff"
 kind = 302
 
+# Glob groups ranges 1
+[[advanced.redirects]]
+source = "/[0-9]/[a-z]/{*}/"
+destination = "http://localhost/new-range/$1/"
+kind = 301
+
+# Glob groups ranges 2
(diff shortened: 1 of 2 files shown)


tests/fixtures/toml/redirects.toml
```diff
@@ -47,6 +47,19 @@ source = "**/{*}.{ttf,otf,woff}"
...
 
+# Glob groups ranges 1
+[[advanced.redirects]]
+source = "/[0-9]/[a-z]/{*}/"
+destination = "http://localhost/new-range/$1/"
+kind = 301
+
+# Glob groups ranges 2
+[[advanced.redirects]]
+source = "/crop-{[!0-9]}/{*}.{png,gif}"
```
mschoettle, line 58: not sure how useful this is. I am struggling to get something like this working where more than one character can be matched. `[!0-9]*` results in a `[^0-9].*` regex because `globset` replaces `*` with `.*`.

joseluisq: I understand. It was just an idea, but we could skip it now if it gives us headaches.

joseluisq (approved):
Good to me. We can always improve our tests over time.

PR #504, opened 2024-11-29
Improve docs configuration and fix anchor links

mschoettle:
## Description

Resolves all info/warnings output by `mkdocs`.

Adds `--strict` to the deploy command. Validation by mkdocs set to `warn`.

## Motivation and Context

Building the documentation site showed some issues with non-existent anchors. `mkdocs` can treat such things as warnings which makes it more visible and building with `--strict` can fail the build process. This helps find issues quicker.

## How Has This Been Tested?

`mkdocs build --strict`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -293,7 +293,7 @@ docs-deploy:
 		-v $(PWD)/docs/content:/docs/docs/content \
 		-v $(PWD)/docs/mkdocs.yml:/docs/mkdocs.yml \
 		-v /tmp/docs:/tmp/docs \
-			static-web-server-docs mkdocs build --strict
+			static-web-server-docs mkdocs build
 	@git checkout gh-pages
 	@git clean -fdx
 	@rm -rf docs/
--- a/Makefile
+++ b/Makefile
@@ -293,7 +293,7 @@ docs-deploy:
 		-v $(PWD)/docs/content:/docs/docs/content \
 		-v $(PWD)/docs/mkdocs.yml:/docs/mkdocs.yml \
 		-v /tmp/docs:/tmp/docs \
-			static-web-server-docs mkdocs build
+			static-web-server-docs mkdocs build --strict
 	@git checkout gh-pages
 	@git clean -fdx
 	@rm -rf docs/
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -3,13 +3,13 @@ site_name: Static Web Server
 site_description: 'A cross-platform, high-performance and asynchronous web server for static files-serving ⚡'
 site_author: 'Jose Quintana'
 site_url: https://static-web-server.net
-dev_addr: 0.0.0.0:8000
 
 # Repository
 repo_name: static-web-server/static-web-server
 repo_url: https://github.com/static-web-server/static-web-server/
 edit_uri: edit/master/docs/content/
 
+strict: true
 docs_dir: 'docs/content'
 site_dir: '/tmp/docs'
 extra_css:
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -293,7 +293,7 @@ docs-deploy:
 		-v $(PWD)/docs/content:/docs/docs/content \
 		-v $(PWD)/docs/mkdocs.yml:/docs/mkdocs.yml \
 		-v /tmp/docs:/tmp/docs \
-			static-web-server-docs mkdocs build
+			static-web-server-docs mkdocs build --strict
```
mschoettle: I wanted to add it to the pipeline as well but there does not seem to be a job/step to build the documentation site. Which of the workflows would be appropriate to add this?

joseluisq: There is no pipeline for this. There is a Makefile task called `docs-deploy` which helps me to automate that process that I have to trigger manually.

It should be useful to leave CI to do the job for us.

docs/mkdocs.yml
joseluisq, line 1: Due to the strict mode now I'm facing this issue. 

```
[+] Running 1/0
 ✔ Container static-web-server-docs  Created                                                                                                                                              0.0s
Attaching to static-web-server-docs
static-web-server-docs  | WARNING -  Config value 'dev_addr': The use of the IP address '0.0.0.0' suggests a production environment or the use of a proxy to connect to the MkDocs server. However, the MkDocs' server is intended for local development purposes only. Please use a third party production-ready server instead.
static-web-server-docs  |
static-web-server-docs  | Aborted with 1 configuration warnings in 'strict' mode!
static-web-server-docs exited with code 1
```

Can you please opt-in the strict mode directly in the config file to be applied for `docs-dev` (development server) and `docs-deploy` (build/publish) tasks?

`docs/mkdocs.yml`

```diff
- dev_addr: 0.0.0.0:8000
+ dev_addr: 127.0.0.1:8000
+ strict: true
```

joseluisq (changes requested):
I like the idea. I just found a minor improvement to make.

Makefile
mschoettle: I meant that it would be good to run `mkdocs build --strict` on every PR to ensure that changes don't break anything. 

Releasing in the pipeline would be helpful as well. Another thing is that it could be helpful to release an image from master with a specific tag. That would allow someone to run the latest state, could be helpful to test it out before a release.

joseluisq: > I meant that it would be good to run `mkdocs build --strict` on every PR to ensure that changes don't break anything.

Yes, a `docs` build check step on `devel` CI could be an option.

> Releasing in the pipeline would be helpful as well.

For releasing, I was thinking about using some extra tool like `release-plz` or similar.

> Another thing is that it could be helpful to release an image from master with a specific tag. That would allow someone to run the latest state, could be helpful to test it out before a release.

For that, a `devel` Docker image built by CI should be useful containing always the latest `master` changes.

mschoettle: > Yes, a docs build check step on devel CI could be an option.

Added.

> For releasing, I was thinking about using some extra tool like release-plz or similar.

Interesting. Thanks for mentioning this tool. Through that I discovered `git-cliff` which looks quite helpful for some projects of mine.

docs/mkdocs.yml
mschoettle: I removed `dev_addr` since it defaults to `0.0.0.0:8000`:

```output
INFO    -  Building documentation...
INFO    -  Cleaning site directory
INFO    -  Documentation built in 2.48 seconds
INFO    -  [14:15:47] Watching paths for changes: 'docs/content', 'mkdocs.yml'
INFO    -  [14:15:47] Serving on http://0.0.0.0:8000/
```

joseluisq (approved):
Thanks!

PR #506, opened 2024-12-01
Prevent single wildcards from matching path separators for URL Rewrites

mschoettle:
## Description

Following the change for URL redirects in #501 make it consistent for URL Redirects as well.

## Related Issue

#501 

## Motivation and Context

A single `*` wildcard should not match the path separator.

## How Has This Been Tested?

`cargo test`


joseluisq (approved):
Thanks!

PR #511, opened 2024-12-10
The closing body tag has been moved right before the closing html tag

alxv-su:
It was a mistake in the html hierarhy: the body contained only h1 tag.

## Description
I moved closing body tag right before closing html tag.

## Related Issue
It is too obvious to create an issue.

## Motivation and Context
Compliance with HTML specification.

## How Has This Been Tested?
I checked source code of directory listing returned by fixed static-web-server using https://validator.w3.org/nu/#textarea

### DIFF
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -411,71 +411,71 @@ fn html_auto_index<'a>(
                 h1 {
                     "Index of " (current_path)
                 }
-            }
-            p {
-                small {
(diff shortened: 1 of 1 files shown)


joseluisq (changes requested):
Good catch! Please just resolve the minor formatting suggestions.

joseluisq (approved):
Thanks!

PR #512, opened 2024-12-10
Build and push `devel` Docker image on master changes via CI workflow

joseluisq:
## Description

This PR refines the work done on https://github.com/static-web-server/static-web-server/pull/508. Basically, it uses a job matrix strategy for the `debian` and `scratch` Docker image variants and makes use of an `x86_64-unknown-linux-musl` static binary for both images to make it work on `scratch`.

**Note:** For now, this workflow only supports one arch/target per Docker variant. However, If multiple targets are needed then it is necessary to refactor the workflow and the 'devel' `Dockerfile` should be adjusted accordingly to be multi-arch.

**Images published only to Github Container Registry:**

https://github.com/static-web-server/static-web-server/pkgs/container/static-web-server

- `ghcr.io/static-web-server/static-web-server:devel` (scratch)
- `ghcr.io/static-web-server/static-web-server:devel-debian` (debian)

## Motivation and Context

Improve the work done on https://github.com/static-web-server/static-web-server/pull/508

## How Has This Been Tested?

- See workflow in action: https://github.com/static-web-server/static-web-server/actions/runs/12266184662
- See Docker images published: https://github.com/static-web-server/static-web-server/pkgs/container/static-web-server

### DIFF
--- a/.github/workflows/release.docker.devel.yml
+++ b/.github/workflows/release.docker.devel.yml
@@ -3,7 +3,7 @@ name: release-docker-devel
 on:
   push:
     branches:
-      - master-ci
+      - master
     paths:
       - .github/workflows/release.docker.devel.yml
       - .cargo/config.toml
--- a/.github/workflows/release.docker.devel.yml
+++ b/.github/workflows/release.docker.devel.yml
@@ -1,0 +1,116 @@
+name: release-docker-devel
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


joseluisq:
The `release-docker-devel` workflow was successful after the merge. See https://github.com/static-web-server/static-web-server/actions/runs/12266415320

Issue #520, opened 2025-01-31
Improvements to X-Forwarded-For is logging and support for X-Real-IP header


dctaf:
### Search for duplicate feature request

- [x] I already searched, and this feature request or improvement is not a duplicate.

### Feature scope

Improve existing functionality

### Feature request related to a problem

While sws can log X-Forwarded-For headers with `--log-forwarded-for`, they aren't always reliable for security because clients and intermediate proxies can set the value to whatever they want. 

Adding support for X-Real-IP would be useful since it's commonly used to log actual end-user IP addresses. 

Also, the current logging misrepresents X-Forwarded-For addresses as ["real"](https://github.com/static-web-server/static-web-server/blob/master/src/log_addr.rs#L55). The header should generally be treated as untrusted user input, but "real" suggests otherwise.

### Describe the solution you'd like

1. Change the existing log message `real_remote_ip` text to `x_forwarded_for`
1. Add a --log-x-real-ip flag that, when enabled, causes --log-remote-addr to use X-Real-IP (if present) instead of the socket address.

### Describe alternatives you've considered

Adding a rich structured JSON logger for requests would be a reasonable alternative.

### Build target

All targets

### Additional context

Related: #494, #495, #112

joseluisq:
It sounds good to me. PR welcome.

joseluisq:
Closing as resolved by #521

MrRinkana:
Note that the "correct" (standardized) header for this purpose is the [forwarded header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded). Or as the haproxy docs put it: "the forwarded header supercedes the X-Forwarded-For header". 

What I'm confused about is this:

> Also, the current logging misrepresents X-Forwarded-For addresses as ["real"](https://github.com/static-web-server/static-web-server/blob/master/src/log_addr.rs#L55). The header should generally be treated as untrusted user input, but "real" suggests otherwise

If the header comes from a trusted proxy, it _**is**_ containing the real ip of the client? And a client can set both the x-real-ip and x-forwarded-for headers in its own request just the same, I don't see how x-real-ip is any more real than the x-forwarded-for (in both cases any webserver or proxy should drop the header unless it's from a trusted proxy).. Could you expand on what you mean @dctaf? 

Also the X-Forwarded-For header is the current de-facto standard way to provide client real ip, X-Real-IP is used much, much less, it's not even mentioned in the MDN docs for example.

[...]

PR #521, opened 2025-02-02
Optionally log X-Real-IP header.

dctaf:
## Description

Adds a --log-x-real-ip option and documentation to enable logging of the
X-Real-IP header in a manner similar to the logging of the X-Forwarded-For
header (and --log-forwarded-for) header. --trustied-proxies is respected.

## Related Issue

#520

## Motivation and Context

For various reasons, many infrastructures use X-Real-IP instead of X-Forwarded-For, and it is useful to have the X-Real-IP value logged.

## How Has This Been Tested?

Manually.

### DIFF
--- a/docs/content/configuration/command-line-arguments.md
+++ b/docs/content/configuration/command-line-arguments.md
@@ -9,7 +9,6 @@ The server can be configured via the following command-line arguments.
 
 ```
 $ static-web-server -h
-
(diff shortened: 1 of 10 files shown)


joseluisq (approved):
Looks fine to me.
Good that you clarified the `X-Forwarded-For` too.

Thanks!

PR #535, opened 2025-03-30
Update dependencies 30-03-2025

joseluisq:
## Description

This PR updates dependencies as usual to provide security, bug fixes and improvements. It is a follow-up for https://github.com/static-web-server/static-web-server/pull/532 that re-enables the ` experimental` cargo feature as the upstream dependency was patched.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,9 +1,1 @@
 [advisories]
-ignore = [
-    # NOTE:
-    # Ignored and disabled from the `static-web-server` final binary (experimental feature) temporarily.
-    # See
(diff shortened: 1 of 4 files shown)


joseluisq:
Discarded for now since it requires bumping up the MSRV to 1.81 or newer.

```
Downloaded tokio-metrics-collector v0.3.1
error: package `prometheus v0.14.0` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.76.0
Either upgrade to rustc 1.81 or newer, or use
cargo update prometheus@0.14.0 --precise ver
where `ver` is the latest version of `prometheus` supporting rustc 1.76.0
Error: Process completed with exit code 101.
```

PR #544, opened 2025-05-21
Add support for downloading a directory as a compressed tarball

### LINKED ISSUE #67: Ability to download zip, tar etc?
As I store directories it would be nice if we can download the whole folder using zip or tar. I think it's a standard functionality for a static web server?

ekangmonyet:
## Description
Support downloading a directory as a compressed tarball.

- Respond with compressed tarball when requested file is a directory and query has `download-archive` parameter set.
- When listing said `<directory>`, add a link with href `?download-archive=1` for user to download all content in directory as compressed tarball (see screenshot).

### Discussion & other notes

- ~~Is the "virtual file" approach appropriate? What should the server do when there is an existing `.tar.gz` file?~~
    - ~~A radical alternative is to introduce an `/api/` group, which may also be used by other non-static-file features such as health endpoint~~
    - ~~Query param seems to be a more versatile and robust solution, e.g. `/<directory>/?download-archive=gzip`~~
    - UPDATE: I feel like query param is the sanest way to go, unless I missed any other caveats or considerations.
- Since this will technically "list directory", if an option is used to toggle this feature, it should respect the directory listing option and possibly warn user/error on conflicting combination. 
- [x] TODO: set `follow_symlinks()` option for tar builder accordingly
- [x] TODO: refactor and cleanup
- [x] TODO: feature flag and runtime options (WIP)
- [x] TODO: test: dir list index - download link follows dir download opt
- [x] TODO: test: dir download - normal, no follow_symlinks
- [x] TODO: verify config file
- [x] TODO: handle ignore hidden file
    - [x] handle recursive append dir ourselves
- [x] TODO: polish testcase 
- ~~TODO: content disposition unicode handling~~ Apparently, non-ASCII is allowed by standard, and coincidentally there is a bug that does not validate the string the way it is documented, so everything works! see: https://github.com/hyperium/http/issues/519

## Related Issue
implement #67 

## Motivation and Context
This feature allows user to download the content of the entire directory all at once.

## How Has This Been Tested?
- Added `download_targz`, `download_targz_no_symlinks` testcase to check downloaded tarball content against directory content
- Added test to ensure download link exists when option is enabled, and ensure it doesnt exists when disabled

(Final implementation)
image

(Previous POC)
image

### Interface of this feature on other system

**Github**
A button in dropdown menu.
image

**Jenkins**
A link below list of files, separated from the list itself
image

### DIFF
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -17,6 +17,9 @@ use std::ffi::{OsStr, OsString};
 use std::io;
 use std::path::Path;
 
+#[cfg(feature = "directory-listing-download")]
+use crate::directory_listing_download::DOWNLOAD_PARAM_KEY;
+
 use crate::{handler::RequestHandlerOpts, http_ext::MethodExt, Context, Result};
 
 /// Non-alphanumeric characters to be percent-encoded
@@ -394,6 +397,15 @@ fn html_auto_index<'a>(
     let sort_attrs = sort_file_entries(entries, order_code);
     let current_path = percent_decode_str(base_path).decode_utf8_lossy();
 
+    #[cfg(feature = "directory-listing-download")]
+    let download_directory_elem = html! {
+        ",  " a href={ "?" (DOWNLOAD_PARAM_KEY) } {
+            "download .tar.gz"
+        }
+    };
+    #[cfg(not(feature = "directory-listing-download"))]
+    let download_directory_elem = html! {};
+
     html! {
         (DOCTYPE)
         html {
@@ -413,11 +425,7 @@ fn html_auto_index<'a>(
                 }
                 p {
                     small {
-                        "directories: " (dirs_count) ", files: " (files_count)
-                        // TODO: check enable dir-archive feature/opt
-                        "  " a href="?download-archive=1" {
-                            "(download all as .tar.gz)"
-                        }
+                        "directories: " (dirs_count) ", files: " (files_count) (download_directory_elem)
                     }
                 }
                 hr;
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -17,6 +17,7 @@ use std::ffi::{OsStr, OsString};
 use std::io;
 use std::path::Path;
 
+use crate::directory_listing_download::DirDownloadFmt;
 #[cfg(feature = "directory-listing-download")]
 use crate::directory_listing_download::DOWNLOAD_PARAM_KEY;
 
@@ -55,6 +56,9 @@ pub struct DirListOpts<'a> {
     pub dir_listing_order: u8,
     /// Directory listing format.
     pub dir_listing_format: &'a DirListFmt,
+    #[cfg(feature = "directory-listing-download")]
+    /// Directory listing download.
+    pub dir_listing_download: &'a Vec<DirDownloadFmt>,
     /// Ignore hidden files (dotfiles).
     pub ignore_hidden_files: bool,
     /// Prevent following symlinks for files and directories.
@@ -99,6 +103,8 @@ pub fn auto_index(opts: DirListOpts<'_>) -> Result<Response<Body>, StatusCode> {
                 content_format: opts.dir_listing_format,
                 ignore_hidden_files: opts.ignore_hidden_files,
                 disable_symlinks: opts.disable_symlinks,
+                #[cfg(feature = "directory-listing-download")]
+                download: opts.dir_listing_download,
             };
             match read_dir_entries(dir_opts) {
                 Ok(resp) => Ok(resp),
@@ -187,6 +193,8 @@ struct DirEntryOpts<'a> {
     content_format: &'a DirListFmt,
     ignore_hidden_files: bool,
     disable_symlinks: bool,
+    #[cfg(feature = "directory-listing-download")]
+    download: &'a Vec<DirDownloadFmt>,
 }
 
 /// It reads a list of directory entries and create an index page content.
@@ -340,6 +348,7 @@ fn read_dir_entries(mut opt: DirEntryOpts<'_>) -> Result<Response<Body>> {
                 files_count,
                 &mut file_entries,
                 opt.order_code,
+                opt.download,
             )
         }
     };
@@ -391,6 +400,7 @@ fn html_auto_index<'a>(
     files_count: usize,
     entries: &'a mut [FileEntry],
     order_code: u8,
+    #[cfg(feature = "directory-listing-download")] download: &'a [DirDownloadFmt],
 ) -> String {
     use maud::{html, DOCTYPE};
 
@@ -398,10 +408,13 @@ fn html_auto_index<'a>(
     let current_path = percent_decode_str(base_path).decode_utf8_lossy();
 
     #[cfg(feature = "directory-listing-download")]
-    let download_directory_elem = html! {
-        ",  " a href={ "?" (DOWNLOAD_PARAM_KEY) } {
-            "download .tar.gz"
-        }
+    let download_directory_elem = match download.is_empty() {
+        true => html! {},
+        false => html! {
+            ",  " a href={ "?" (DOWNLOAD_PARAM_KEY) } {
+                "download .tar.gz"
+            }
+        },
     };
     #[cfg(not(feature = "directory-listing-download"))]
     let download_directory_elem = html! {};
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -294,16 +294,6 @@ fn read_dir_entries(mut opt: DirEntryOpts<'_>) -> Result<Response<Body>> {
         file_entries.push(entry);
     }
 
-    // TODO: check enable dir-archive feature
-    // TODO: make this post process
-    file_entries.push(FileEntry {
-        name: "(download as .tar.gz)".into(),
-        mtime: None,
-        size: None,
-        r#type: FileType::File, // TODO: do we need special handling?
-        uri: "?download-archive=1".into(),
-    });
-
     // Check the query request uri for a sorting type. E.g https://blah/?sort=5
     if let Some(q) = opt.uri_query {
         let mut parts = form_urlencoded::parse(q.as_bytes());
@@ -424,6 +414,10 @@ fn html_auto_index<'a>(
                 p {
                     small {
                         "directories: " (dirs_count) ", files: " (files_count)
+                        // TODO: check enable dir-archive feature/opt
+                        "  " a href="?download-archive=1" {
+                            "(download all as .tar.gz)"
+                        }
                     }
                 }
                 hr;
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -295,12 +295,13 @@ fn read_dir_entries(mut opt: DirEntryOpts<'_>) -> Result<Response<Body>> {
     }
 
     // TODO: check enable dir-archive feature
+    // TODO: make this post process
     file_entries.push(FileEntry {
         name: "(download as .tar.gz)".into(),
         mtime: None,
         size: None,
         r#type: FileType::File, // TODO: do we need special handling?
-        uri: ".tar.gz".into(),
+        uri: "?download-archive=1".into(),
     });
 
     // Check the query request uri for a sorting type. E.g https://blah/?sort=5
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -294,6 +294,15 @@ fn read_dir_entries(mut opt: DirEntryOpts<'_>) -> Result<Response<Body>> {
         file_entries.push(entry);
     }
 
+    // TODO: check enable dir-archive feature
+    file_entries.push(FileEntry {
+        name: "(download as .tar.gz)".into(),
+        mtime: None,
+        size: None,
+        r#type: FileType::File, // TODO: do we need special handling?
+        uri: ".tar.gz".into(),
+    });
+
     // Check the query request uri for a sorting type. E.g https://blah/?sort=5
     if let Some(q) = opt.uri_query {
         let mut parts = form_urlencoded::parse(q.as_bytes());
--- a/src/directory_listing.rs
+++ b/src/directory_listing.rs
@@ -411,7 +411,7 @@ fn html_auto_index<'a>(
     let download_directory_elem = match download.is_empty() {
         true => html! {},
         false => html! {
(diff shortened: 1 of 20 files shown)


ekangmonyet:
This is an early draft of the feature, implementing a prototype for #67.

Disclaimer: I do not have much experience in async Rust (or even that much experience in Rust in general), my implementation might be subpar/unidiomatic. All kind of feedback is very much appreciated.

joseluisq:
Sorry for the delay, here are a few comments on the feature. I will add more if needed during the next days.
Regarding the code, I will comment on that directly.

But let me know what your thoughts are on this.

--

> * UPDATE: I feel like query param is the sanest way to go, unless I missed any other caveats or considerations.

Agree, let's go with a query param for the directory download.
I suggest this query: `/?download` that _will_ be the short version of `/?download=tarball`. Let's support only the first (`/?download`) for now. If we want to add zip or another format in the future then we can just extend the query to be like the second (`/?download=tarball|zipball|etc`).

> * Since this will technically "list directory", if an option is used to toggle this feature, it should respect the directory listing option and possibly warn user/error on conflicting combination.

What do you mean exactly by "conflicting combination"?
The question is, do we want to make this directory download feature dependent on the directory listing, or should this feature be standalone and also complement the directory listing instead (if enabled)? 

For example, I think this feature will be behind an option flag like `directory-download=tarball` (just an idea), but it could work regardless of the directory listing. Of course, if the directory listing is enabled with the download, we have to display the link in the HTML template.

> About the link text on the directory listing.

I believe we can just say "download tarball" or "download tar.gz" whatever fits better. Something like:

image

ekangmonyet:
> Sorry for the delay

Hey, please take your time, don't worry about it! Thanks for the review too!

> I suggest this query: `/?download` that _will_ be the short version of `/?download=tarball`.

Yeap sure, `download` will do just fine. It is not too relevant for now, but does `tarball` always equate to gzipped tar? What if we want to support XZ or bzip2?

> What do you mean exactly by "conflicting combination"?

IMO in some cases, `directory-listing` will be treated as a security toggle, i.e. in no way expose the list of content in the directory. Thus, if somehow the user disabled `directory-listing` but enabled `directory-download`, I guess that will be conflicting, in the sense that the content can now be leaked by downloading the tarball?

> I believe we can just say "download tarball" or "download tar.gz" whatever fits better.

Yeap sure, sounds great!

joseluisq:
> Yeap sure, `download` will do just fine. It is not too relevant for now, but does `tarball` always equate to gzipped tar? What if we want to support XZ or bzip2?

Good question, I was thinking of only supporting in the future `tar.gz` and `zip` as those two are the popular ones (Unix/Windows).
To be consistent, let's forget the tarball or zipball thing and use the extension as a reference.
We will use `/?download` (which means `tar.gz`) and later expand it to `/?download=tar.gz` or `/?download=zip` if needed in the future.

The text link can also be `download tar.gz`.

> IMO in some cases, `directory-listing` will be treated as a security toggle, i.e. in no way expose the list of content in the directory. Thus, if somehow the user disabled `directory-listing` but enabled `directory-download`, I guess that will be conflicting, in the sense that the content can now be leaked by downloading the tarball?

I got it. In that case, let's make it dependent on `directory-listing=true`, so if a user needs download functionality, the user has to enable it like this:

```sh
$ static-web-server -p 1234 -d ./public \
    --directory-listing --directory-listing-download=tar.gz
```

The `--directory-listing-download` option should be empty by default (which means it is disabled).
This will also allow us later if we need to support `zip`, to just expand it to a list like `--directory-listing-download=tar.gz,zip`.

src/static_files.rs
```diff
@@ -184,6 +231,66 @@ pub async fn handle(opts: &HandleOpts<'_>) -> Result<StaticFileResponse, StatusC
...
+                }
+                return Ok(StaticFileResponse {
+                    resp,
+                    file_path: resp_file_path,
+                });
+            } else {
+                panic!("unexpected error {}", file_path.to_string_lossy());
+            }
+        }
+    }
```
joseluisq: Here are a few minor suggestions:
- Consider moving the code to a `src/directory_listing_download.rs` file and use those functions in here accordingly.
- Handle any panic or unwrap by returning the proper error responses.

src/static_files.rs
```diff
@@ -40,6 +47,46 @@ use crate::{
     directory_listing::{DirListFmt, DirListOpts},
```
joseluisq, line 40: Once the logic is moved out of this file, then import the module here under the `feature = "directory-listing"` cargo feature.

joseluisq (changes requested):
That's for now, I believe you are still refactoring the code and adding the SWS feature option. SWS options can be added via [`src/settings/cli.rs`](https://github.com/static-web-server/static-web-server/blob/master/src/settings/cli.rs) and [`src/settings/file.rs`](https://github.com/static-web-server/static-web-server/blob/master/src/settings/file.rs).

So I will wait for your changes first.
Just let me know if you have questions.

ekangmonyet:
I think it's almost ready for review, just a few questions/confirmations:

- This feature should probably respect the `ignore_hidden_files` option as well, to implement that I think we have to replace the `append_dir_all` with our own implementation. Do we want to do this in this commit as well?
- I tried dancing with clap and the current state is the best I can do:
    - `--directory-listing-download=targz` to enable
    - `--directory-listing-download=none` to disable (mixing `none` with other non-`none` values will panic, should we handle this differently as well?, I can't find a way to do it all in clap, yet)
    - future: `--directory-listing-download=targz,zip,...`
    - this flag requires `-z` (or the long one)
- on error handling:
    - I think we can still try our best to respond if error occurs when making the header (most likely due to code changes/upstream behavior changes), instead of responding with 500?
    - For compression task, the only way to signal error is by calling `sender.abort()`, which IMO doesn't do much anyway. The most likely error seems to be due to broken pipe, i.e. user aborted the download midway. In that situation, there is nothing we can do since the session has already been closed anyway.

joseluisq:
> * This feature should probably respect the `ignore_hidden_files` option as well, to implement that I think we have to replace the `append_dir_all` with our own implementation. Do we want to do this in this commit as well?

Yes, since `ignore_hidden_files` skips dotfiles, we have to implement it. It may not be a big deal to implement it. See https://docs.rs/async-tar/latest/src/async_tar/builder.rs.html#585.

> * I tried dancing with clap and the current state is the best I can do:
>   * `--directory-listing-download=targz` to enable
>   * `--directory-listing-download=none` to disable (mixing `none` with other non-`none` values will panic, should we handle this differently as well?, I can't find a way to do it all in clap, yet)
>   * future: `--directory-listing-download=targz,zip,...`
>   * this flag requires `-z` (or the long one)

What if we represent the `directory_listing_download` as `Vec<Option<DirDownloadFmt>>` and remove the enum's `None` case of `DirDownloadFmt` (not tested)?

I will investigate from my side as well. It would be much better to support an enum, but if not possible, we could use a simple `String` instead, where an empty string means disabled. Remember that if we add `zip` archive support, we could use both, like `--directory-listing-download=targz,zip`. If we use a string, then only check for `s.is_empty()` or `s == "targz"`.
In the future, we could read just the string value and split it by a comma separator.

> * on error handling:
>   
>   * I think we can still try our best to respond if error occurs when making the header (most likely due to code changes/upstream behavior changes), instead of responding with 500?
>   * For compression task, the only way to signal error is by calling `sender.abort()`, which IMO doesn't do much anyway. The most likely error seems to be due to broken pipe, i.e. user aborted the download midway. In that situation, there is nothing we can do since the session has already been closed anyway.

This sounds fine to me. If we do not panic the server thread, we are fine.

ekangmonyet:
> What if we represent the `directory_listing_download` as `Vec<Option<DirDownloadFmt>>` and remove the enum's `None` case of `DirDownloadFmt` (not tested)?

Do we absolutely need to allow explicitly disabling using `--directory-listing-download=`, or will the lack of said option be sufficient? If that's the case, `None` can be removed.

src/directory_listing.rs
```diff
@@ -388,12 +400,25 @@ fn html_auto_index<'a>(
...
     use maud::{html, DOCTYPE};
 
     let sort_attrs = sort_file_entries(entries, order_code);
     let current_path = percent_decode_str(base_path).decode_utf8_lossy();
 
+    #[cfg(feature = "directory-listing-download")]
+    let download_directory_elem = match download.is_empty() {
+        true => html! {},
+        false => html! {
+            ",  " a href={ "?" (DOWNLOAD_PARAM_KEY) } {
```
joseluisq:
```suggestion
            ", " a href={ "?" (DOWNLOAD_PARAM_KEY) } {
```

src/directory_listing_download.rs
```diff
@@ -0,0 +1,200 @@
...
+pub struct ChannelBuffer {
+    s: Sender,
+}
+
+impl tokio::io::AsyncWrite for ChannelBuffer {
+    fn poll_write(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &[u8],
+    ) -> std::task::Poll<std::result::Result<usize, std::io::Error>> {
```
joseluisq:
```suggestion
    ) -> std::task::Poll<Result<usize, std::io::Error>> {
```

src/directory_listing_download.rs
```diff
@@ -0,0 +1,200 @@
...
+                Err(e) => Ready(Err(io::Error::new(io::ErrorKind::BrokenPipe, e))),
+            },
+            Pending => Pending,
+        }
+    }
+
+    fn poll_flush(
+        self: std::pin::Pin<&mut Self>,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<std::result::Result<(), std::io::Error>> {
```
joseluisq:
```suggestion
    ) -> std::task::Poll<Result<(), std::io::Error>> {
```

src/directory_listing_download.rs
```diff
@@ -0,0 +1,200 @@
...
+        std::task::Poll::Ready(Ok(()))
+    }
+
+    fn poll_shutdown(
+        self: std::pin::Pin<&mut Self>,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<std::result::Result<(), std::io::Error>> {
```
joseluisq:
```suggestion
    ) -> std::task::Poll<Result<(), std::io::Error>> {
```

src/directory_listing_download.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    // NOTE: Since it is not possible to handle error gracefully, we will
+    // just stop writing when error occurs. It is also not possible to call
+    // sender.abort() as it is protected behind the Builder to ensure
+    // finish() is successfully called.
+
+    // adapted from async_tar::Builder::append_dir_all
+    let mut stack = vec![(src_path.to_path_buf(), true, false)];
+    while let Some((src, is_dir, is_symlink)) = stack.pop() {
+        let dest = path.join(src.strip_prefix(&src_path).unwrap());
```
joseluisq:
```suggestion
        let dest = path.join(src.strip_prefix(&src_path)?);
```

src/directory_listing_download.rs
```diff
@@ -0,0 +1,200 @@
...
+    );
+    match HeaderValue::from_str(hvals.as_str()) {
+        Ok(hval) => {
+            resp.headers_mut()
+                .insert(hyper::header::CONTENT_DISPOSITION, hval);
+        }
+        Err(err) => {
+            // not fatal, most browser is able to handle the download since
+            // content-type is set
+            tracing::error!("cant make content disposition from {}: {:?}", hvals, err);
```
joseluisq:
```suggestion
            tracing::error!("can't make content disposition from {}: {:?}", hvals, err);
```

joseluisq (changes requested):
I did a quick review and encountered a few minor changes. I will continue reviewing and testing in a second round.

src/settings/cli.rs
joseluisq, line 1: Since this is a new feature, we have to document it.

For that, let's update the [directory listing page](https://github.com/static-web-server/static-web-server/blob/master/docs/content/features/directory-listing.md), adding a new entry called `Directory Download` below `Output format`.

Basically, a short description of the feature and an example of how to use it should be enough.

joseluisq (changes requested):

src/directory_listing_download.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/// Initializes directory listing download
+pub fn init(formats: &Vec<DirDownloadFmt>, handler_opts: &mut RequestHandlerOpts) {
+    for fmt in formats {
+        // Use naive implementation since the list is not expected to be long
+        if !handler_opts.dir_listing_download.contains(fmt) {
+            handler_opts.dir_listing_download.push(fmt.to_owned());
+        }
+    }
```
joseluisq, line 66: It should not be bad to add an info log entry like `tracing::info!("directory listing download: enabled={enabled}");`.

src/static_files.rs
```diff
@@ -179,6 +188,40 @@ pub async fn handle(opts: &HandleOpts<'_>) -> Result<StaticFileResponse, StatusC
         });
     }
 
+    // Directory listing download
+    // Check if "directory listing download" feature is enabled,
+    // if current path is a valid directory and
+    // if query string has parameter "download" set
+    #[cfg(feature = "directory-listing-download")]
+    if is_dir && !opts.dir_listing_download.is_empty() {
```
joseluisq: Right now, if we want to download a dir via a `/?download` request and if that dir already contains an `index.html` file, the dir archive is downloaded anyway. Perhaps we should still serve only the `index.html` file and skip any archive downloading.

I believe, if the download feature depends on the directory listing, then we should move this code down to line 231. Thoughts?

image

joseluisq (changes requested):
There are a few last minor suggestions. It looks fine to me overall.

src/static_files.rs
ekangmonyet: Yeap, sounds about right!

src/directory_listing_download.rs
ekangmonyet: EDIT: yeap I added a line to be consistent with the rest of the features

Do we want to print if it's disabled?

joseluisq: That's fine

joseluisq (approved):
Looks fine to me. Ready to be available in the upcoming release.
Thanks!

PR #551, opened 2025-07-01
Prefer a less-generic config file `sws.toml` as default and keep the old `config.toml` temporarily

### LINKED ISSUE #550: Use a less generic default config file name

### Feature scope

The default `config.toml` is the same as for other projects, such as Zola for example. 

It could be great to have `sws_config.toml` or `sws.toml` as default configuration name instead, to avoid conflicts.

Using other static web server as default for deployment to avoid conflicts

### Build target

All targets

### Additional context

_No response_

davlgd:
This PR tends to fix #550 

## Description
It changes default config file name to `sws_config.toml` to avoid conflicts. Docs/tests are updated accordingly, if you prefer I can create dedicated commits for that.

## Related Issue
#550 

## Motivation and Context
We use Zola and potentially other projects already using a `config.toml` file as default. When we launch the project, there is unwanted logs as sws use the zola config.toml as its own. 

Another solution could be to not read/use `config.toml` file not valid for SWS, but I think  it's better to avoid such generic file name. And the sooner the better on such change.

Feel free to close this if you prefer to continue with the current file name and go for another solution. 

## How Has This Been Tested?
I've build the project and used it with a `sws_config.toml` without env var or flag to point it. It was perfectly loaded and used.


joseluisq:
Hey @davlgd, based on https://github.com/static-web-server/static-web-server/issues/550#issuecomment-3023959411.

Here are some steps you can take:

- Use `sws.toml` as the default value in the `cli.rs` file.
- Check for the old `config.toml`  file at start-up, if it exists, then use that config.
- Otherwise, if the old `config.toml` does not exist, then check for `sws.toml` (default behaviour).

Let me know your thoughts on that.

davlgd:
@joseluisq I've made the change as asked. Feel free to update warning message

src/settings/mod.rs
```diff
@@ -224,9 +224,15 @@ impl Settings {
...
-        if let Some((settings, config_file_resolved)) = read_file_settings(&opts.config_file)? {
+        let to_use_config_file = match Path::new("./config.toml").is_file() {
+            true => {
+                eprintln!("Deprecated: 'config.toml' found, rename it to 'sws.toml' to prepare for future releases");
+                PathBuf::from("./config.toml")
+            }
+            false => opts.config_file.clone(),
+        };
+
+        if let Some((settings, config_file_resolved)) = read_file_settings(&to_use_config_file)? {
```
joseluisq, line 235: Just a comment for me to remove that check later in a future release.

joseluisq (approved):
Looks fine to me.
I will take care of the CI warnings.

Thanks!

PR #570, opened 2025-10-13
Virtual Hosts: support HTTP/2 using :authority

### LINKED ISSUE #569: Virtual Hosts don't work with HTTP/2

- [x] I already searched, and this issue is not a duplicate.

### Issue scope

BSD (E.g. FreeBSD, NetBSD, illumos)

### Describe the bug

I noticed that virtual hosts are not working for me. I found out that this is because the HTTP host header is used to determine the hostname that is being accessed, and my browser does not send that header. Turns out HTTP/2 uses a different mechanism to send this information, the ":authority" pseudo-header field.

### How to reproduce it

Try to use virtual hosts with HTTP/2

### Expected behavior

Virtual hosts should work with HTTP/2

### Complementary information

_No response_

### Build target

aarch64-unknown-linux-gnu

### Environment and specs

Comeon, this is irrelevant. I already made a fix but am not allowed to create a PR without creating an issue first.

### Additional context

_No response_

CrazyCraftix:
It resolves #569

## Description
So far, the HTTP host header has been used to determine the hostname that is being accessed. In HTTP/2, this header is not guaranteed to be sent. Instead, the ":authority" pseudo-header field is used. I added a check for that header to make virtual hosts also work with HTTP/2.

Also I found out that the HTTP host header may also contain the port number, so I added a check to remove it if it's there.

## How Has This Been Tested?
Not thoroughly, just manually for my usecase, so there may still be problems I didn't consider.
I suppose an automated test would be nice but I'm not sure what's the best way to do this.

### DIFF
--- a/docs/content/download-and-install.md
+++ b/docs/content/download-and-install.md
@@ -1,13 +1,13 @@
 <!-- Content generated. DO NOT EDIT. -->
 # Download and Install
 
-Latest **v2.38.0** release `2025-07-21` ([changelog](https://github.com/static-web-server/static-web-server/releases/tag/v2.38.0), [sha256sum](https://github.com/static-web-server/static-web-server/releases/download/v2.38.0/static-web-server-v2.38.0-SHA256SUM))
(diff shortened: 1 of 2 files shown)


joseluisq (changes requested):
I will add some test cases.

joseluisq:
@CrazyCraftix I made the mistake of force-pushing accidentally when wanting to add tests.
I created a new PR https://github.com/static-web-server/static-web-server/pull/571 with your changes and added tests.
I apologize.

CrazyCraftix:
All good, thanks for integrating it!

PR #577, opened 2025-11-04
Content negotiation for Markdown files

davlgd:
## Description

This PR adds a `--accept-markdown` / `SERVER_ACCEPT_MARKDOWN` feature. Once enabled, if a `GET` or `HEAD` request is received for `path/to/file` with `Accept: text/markdown` header, the server checks if a markdown version exists and send it as response. Checked paths are: 
* `path/to/file.md`
* `path/to/file.html.md`
* `path/to/file/index.html.md`

Thus it helps to get an existing markdown file of an HTML page on a website which respects [llms.txt standard](https://llmstxt.org/). It states that:

> Pages on websites that have information that might be useful for LLMs to read provide a clean markdown version of those pages at the same URL as the original page, but with .md appended. (URLs without file names should append index.html.md instead.)

Agents/LLMs can ask for a page with `Accept: text/markdown` and get directly the markdown version. Learn more in added documentation. 

## Motivation and Context
We build our documentation with Hugo which now includes a native llms.txt support. We use SWS as static server but it lacks `Accept` header configuration/support. So I thought about adding this feature to allow a better discovery of our content in markdown format.

## How Has This Been Tested?
I've added tests, tried it with our documentation.


src/handler.rs
```diff
@@ -282,6 +285,18 @@ impl RequestHandler {
...
+            if self.opts.accept_markdown {
+                if let Some(result) =
+                    crate::markdown::pre_process(&self.opts, req, base_path, req.uri().path())
+                {
+                    let resp = result?;
+                    // Post-process the markdown response (apply CORS, etc.)
+                    let resp = cors::post_process(&self.opts, req, resp)?;
+                    return Ok(resp);
+                }
+            }
```
joseluisq: I would suggest instead of handling the file manually in `markdown::pre_process` (trying metadata, reading the whole file, preparing the response, checking CORS, etc), just replace the `uri_path` with the markdown variant and then make sure to pass it over to the `static_files::handle` options.

By doing this, you will let `static_files::handle` do its usual job for you, such as sanitize the path, find the metadata, read the file stream, add proper headers, and prepare the file or error response.
Also, the other post-process functions can be applied later on, like compression, cors, control-headers, security headers, custom headers, etc.

If you need to add extra response headers, you can modify the response later by creating a `post_process` function.

Here is an idea:

```suggestion
            let mut uri_path = req.uri().path();

            // Check for markdown content negotiation (only if enabled)
            if self.opts.accept_markdown {
                if let Some(uri_path_md) =
                    crate::markdown::pre_process(&self.opts, req, base_path, &uri_path)
                {
                    uri_path = uri_path_md;
                }
            }
```

src/markdown.rs
joseluisq, line 1: As I mentioned in my previous [comment](https://github.com/static-web-server/static-web-server/pull/577/files#r2492117225), if you do that, then `src/markdown.rs` can be simplified to just return an `Option<&str>` (markdown URI path variant) instead.

After that, the URI path sanitization, trying the metadata for the file (`HEAD`) or reading the file stream (`GET`) will be done by the `static_files::handle`.

src/markdown.rs
```diff
@@ -0,0 +1,148 @@
...
+            .uri("/test")
+            .body(Body::empty())
+            .unwrap();
+
+        // Without Accept header, should return None (no markdown handling)
+        let base_path = std::path::Path::new("/tmp");
+        let result = pre_process(&opts, &req, base_path, "/test");
+
+        assert!(result.is_none());
+    }
```
joseluisq, line 91: Also, the tests here need to be adjusted.

And regarding testing requests. Please add some integration tests for when enabling and disabling the feature, as well as when making `HEAD` and `GET` requests.
You can guide yourself from existing tests under the [./tests](https://github.com/static-web-server/static-web-server/tree/master/tests) directory (e.g. `tests/handler.rs`). Just create a new test file and a fixture.

joseluisq (changes requested):
I like the idea of content negotiation. In general, this could be extended in the future to support, for example, language and media (e.g., images) negotiation.

I gave the MR a first review. I will take a second look in the coming hours.

If something is not clear or you need help, just let me know.

src/markdown.rs
davlgd: Thanks a lot, I'll update all that

davlgd:
I've updated following your recommandations, updated tests and added integration tests. Let me know if you need any further modification.

src/markdown.rs
```diff
@@ -0,0 +1,144 @@
...
+use std::path::Path;
+
+use crate::{
+    Error, fs::meta::try_markdown_variant, handler::RequestHandlerOpts, headers_ext::Accept,
+};
+
+/// Pre-process a request to check if a markdown variant URI should be used.
+/// Returns the modified URI path if a markdown variant exists, None otherwise.
+pub(crate) fn pre_process<T>(
+    _opts: &RequestHandlerOpts,
```
joseluisq: Since `opts` is not used, you can remove it.

```suggestion
```

src/markdown.rs
```diff
@@ -0,0 +1,144 @@
...
+    let accepts_markdown = req
+        .headers()
+        .typed_get::<Accept>()
+        .map(|accept| accept.accepts_markdown())
+        .unwrap_or(false);
+
+    if !accepts_markdown {
+        return Ok(resp);
+    }
+
```
joseluisq: (1) If I make a `GET` request to fetch an `.html` file, then the markdown headers below are applied.
I suggest simplifying the code to do this below, so we make sure to apply those headers only to the markdown variant if applicable.

```suggestion
pub(crate) fn post_process(
    is_markdown_variant: bool,
    opts: &RequestHandlerOpts,
    mut resp: Response<Body>,
) -> Result<Response<Body>, Error> {
    if !is_markdown_variant || !opts.accept_markdown {
        return Ok(resp);
    }

```

src/handler.rs
```diff
@@ -327,6 +338,9 @@ impl RequestHandler {
             // Append CORS headers if they are present
             let resp = cors::post_process(&self.opts, req, resp)?;
 
+            // Set Content-Type for markdown files
+            let resp = crate::markdown::post_process(&self.opts, req, resp)?;
+
```
joseluisq, line 343: (2) Once `markdown::post_process` function is updated ([comment 1](https://github.com/static-web-server/static-web-server/pull/577/files#r2495496659)), then the caller side here can be changed to:

```suggestion
            let resp = markdown::post_process(uri_path_md.is_some(), &self.opts, resp)?;

```

tests/fixtures/markdown/article.html
joseluisq, line 1: Can we please move the whole `docker/public/markdown` directory to `tests/fixtures/public/markdown`?

tests/markdown.rs
joseluisq, line 1: Please also add a few `GET`/`HEAD` request tests for fetching HTML content when `accept-markdown` is enabled and verify that the headers are correctly applied.

joseluisq (changes requested):
Last changes to make, then we should be ready.

davlgd:
I've added 2 commits to let you check changes more easily, feel free to squash/remessage them

joseluisq (approved):
All fine from my side.
Thanks!

PR #578, opened 2025-11-14
Add `exit_on_error` option for `Server::run_server_on_rt`

frnsys:
add `exit_on_error` option when starting server to choose if the entire process exits if the server fails to start

## Description

Added `exit_on_error` as an additional argument to `run_server_on_rt`. When set to `true` this exits the entire process if the server fails to start (the current behavior). When set to false the error is propagated to the caller of `run_server_on_rt`. `run_standalone` and `run_as_service` don't expose this parameter, passing `true` to keep the current behavior.

## Motivation and Context

I'm using `static-web-server` as a library and found it killed my entire program when it failed to start, but I need a more graceful exit (clean up, etc).

## How Has This Been Tested?

I ran `cargo test` and all tests passed. This was on Linux.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -115,7 +115,7 @@ impl Server {
     /// Build and run the multi-threaded `Server` on the Tokio runtime.
     ///
     /// Setting `exit_on_error` to `true` will exit the entire process if
-    /// the server fails to start.
+    /// the server fails to start (previous behaviour).
     pub fn run_server_on_rt<F>(self, cancel_recv: Option<Receiver<()>>, cancel_fn: F, exit_on_error: bool) -> Result
     where
         F: FnOnce(),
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -109,31 +109,38 @@ impl Server {
...
     {
-        self.run_server_on_rt(cancel, cancel_fn)
+        self.run_server_on_rt(cancel, cancel_fn, true)
     }
 
     /// Build and run the multi-threaded `Server` on the Tokio runtime.
-    pub fn run_server_on_rt<F>(self, cancel_recv: Option<Receiver<()>>, cancel_fn: F) -> Result
+    ///
+    /// Setting `exit_on_error` to `true` will exit the entire process if
+    /// the server fails to start.
```
joseluisq:
```suggestion
    /// Setting `exit_on_error` to `true` will exit the entire process if
    /// the server fails to start (previous behaviour).
```

joseluisq (changes requested):
Good idea, just a small update to make and fix formatting.

joseluisq (changes requested):
There are formatting issues. Fix them by applying `cargo fmt`.

joseluisq (approved):
Thanks!

PR #595, opened 2025-12-04
docs: add missing newline to ensure that bulleted list is rendered correctly

mschoettle:
## Description

I was going through the changelog of the latest version and noticed an incorrectly rendered list in https://static-web-server.net/features/markdown-content-negotiation/#llm-friendly-content-with-llmstxt at the end of that section:

> This approach is superior to traditional .md URL suffixes because: - Uses standard HTTP content negotiation - Same URL serves both HTML (browsers) and Markdown (LLMs) - No need for separate URL structures or redirects - Follows REST principles for resource representation

## Motivation and Context

Note that it would be possible to catch (and fix) this automatically via `mdformat` with the `mdformat-mkdocs` plugin. I use it as a pre-commit hook but could also be added to CI only.

Let me know if this is something you would find useful and I could contribute a PR.

### DIFF
--- a/docs/content/features/markdown-content-negotiation.md
+++ b/docs/content/features/markdown-content-negotiation.md
@@ -71,6 +71,7 @@ curl -H "Accept: text/markdown" https://example.com/docs/
 ```
 
 This approach is superior to traditional `.md` URL suffixes because:
+
 - Uses standard HTTP content negotiation
 - Same URL serves both HTML (browsers) and Markdown (LLMs)
 - No need for separate URL structures or redirects


joseluisq:
Hi, checking and formatting the docs via https://github.com/hukkin/mdformat on CI sounds interesting.

You could use something like this:

```sh
mdformat --check ./*.md docs/**
```

It is enough to check for all root-level only and `docs/**` markdown files.

Can you please add it to CI as a new step? I created a [dedicated workflow](https://github.com/static-web-server/static-web-server/blob/master/.github/workflows/devel.project.docs.yml) for the project docs. Just make sure to add it after the typos check and before the mkdocs build.

--

Btw, I believe you can close this PR and address the same in a new one by applying the formatting via `mdformat ./*.md docs/**`.

mschoettle:
Closing in favour of #597.

PR #597, opened 2025-12-05
Format Markdown files with mdformat

mschoettle:
## Description

Formats all (almost) Markdown files with `mdformat` and the `mdformat-mkdocs` plugin.

Adds `mdformat --check` in a step to the docs CI workflow.

## Related Issue

n/a

## Motivation and Context

As discussed in:

- #595 

## How Has This Been Tested?

Locally and via the the new CI step.

## To consider

1. I added the configuation file (`.mdformat.toml`) because it makes it easier to run `mdformat` in various places with the same config.

    I wanted to add `mdformat` to the Dockerfile, unfortunately, `--exclude` requires Python 3.13` and the `mkdocs-material` image still uses Python 3.11. It could be changed to though if you think that would be a better experience as well (either by using `uv` to run `mdformat` or changing the base image).
    It would then also make it possible to add it to the Makefile.

2. I enabled `--compact-tables` to avoid sweeping formatting changes in the tables but we could remove that.

3. It is also possible to format code blocks. I tried it out with the `mkdocs-beautysh` for shell scripts and `mkdocs-config` for TOML, JSON, and YAML files. Shell scripts didn't fully "work" due to the comments in multiline arguments. It would require a trailing slash at the end of those comments to avoid indentation changes in the next line. Via `beautysh` it helped find two missing trailing slashes :)

### DIFF
--- a/.github/workflows/devel.project.docs.yml
+++ b/.github/workflows/devel.project.docs.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Check Markdown formatting
       run: |
         pip install uv
-        uvx --with mdformat-mkdocs mdformat --number --compact-tables --exclude "docs/content/download-and-install*.md" --exclude "CHANGELOG" --check ./*.md docs/
+        uvx --python ">=3.13" --with mdformat-mkdocs mdformat --check ./*.md docs/
 
     - name: Build docs
       run: |
--- a/.github/workflows/devel.project.docs.yml
+++ b/.github/workflows/devel.project.docs.yml
@@ -24,6 +24,11 @@ jobs:
       with:
         config: ./.github/workflows/config/typos.toml
 
+    - name: Check Markdown formatting
+      run: |
+        pip install uv
+        uvx --with mdformat-mkdocs mdformat --number --compact-tables --exclude "docs/content/download-and-install*.md" --exclude "CHANGELOG" --check ./*.md docs/
+
     - name: Build docs
       run: |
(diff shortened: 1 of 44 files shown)


mschoettle:
Ready! :)

See the "To consider" section for some things to discuss/consider.

joseluisq:
> 1. I added the configuation file (`.mdformat.toml`) because it makes it easier to run `mdformat` in various places with the same config.

It is fine for now. I thought mdformat supports a config file argument to move it to `.github/workflows/config/` but it doesn't.

>    I wanted to add `mdformat` to the Dockerfile, unfortunately, `--exclude` requires Python 3.13`and the`mkdocs-material`image still uses Python 3.11. It could be changed to though if you think that would be a better experience as well (either by using`uv`to run`mdformat` or changing the base image).
>    It would then also make it possible to add it to the Makefile.

We could improve `Dockerfile` and build `mkdocs-material` ourselves later, feel free.

> 2. I enabled `--compact-tables` to avoid sweeping formatting changes in the tables but we could remove that.
> 3. It is also possible to format code blocks. I tried it out with the `mkdocs-beautysh` for shell scripts and `mkdocs-config` for TOML, JSON, and YAML files. Shell scripts didn't fully "work" due to the comments in multiline arguments. It would require a trailing slash at the end of those comments to avoid indentation changes in the next line. Via `beautysh` it helped find two missing trailing slashes :)

It looks fine to me for now.

joseluisq (approved):
Thanks!

PR #606, opened 2025-12-26
Prevent hidden root path to be ignored by `--ignore-hidden-files` feature

mightyiam:
## Description

This PR fixes an issue where SWS returned 404 when serving from a hidden directory (i.e., `/.root`) together with the `--ignore-hidden-files` feature being enabled.

**Server configuration**

```sh
$ static-web-server -p 8788 -d .public/ --ignore-hidden-files
```

**Before**

```sh
$ curl -I http://localhost:8788
# HTTP/1.1 404 Not Found
# content-type: text/html; charset=utf-8
# content-length: 312
# accept-ranges: bytes
# vary: accept-encoding
# cache-control: max-age=86400
# date: Fri, 26 Dec 2025 14:24:54 GMT
```

**After (expected)**

```sh
$ curl -I http://localhost:8788
# HTTP/1.1 200 OK
# content-length: 639
# content-type: text/html
# accept-ranges: bytes
# last-modified: Sat, 20 Dec 2025 11:48:14 GMT
# vary: accept-encoding
# cache-control: max-age=86400
# date: Fri, 26 Dec 2025 14:27:05 GMT
```

## Motivation and Context

I think that only components of the path _under_ the base path should be considered for being hidden.
The base path is explicit and there is no point in considering its components.
Current behavior was undesirable when I was trying to make automated tests that make use of this project.
In my tests, I have created temporary base directories that happen to have a hidden component.

## How Has This Been Tested?

Added a test.

### DIFF
--- a/tests/fixtures/.hidden-root/foo.html
+++ b/tests/fixtures/.hidden-root/foo.html
@@ -1,0 +1,1 @@
+<h1>Just an HTML file</h1>
--- a/tests/static_files.rs
+++ b/tests/static_files.rs
@@ -1520,6 +1520,99 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn hidden_base_path_not_ignored() {
+        let root_dir = PathBuf::from("tests/fixtures/.hidden-root");
+        let headers = HeaderMap::new();
+
+        for method in [Method::HEAD, Method::GET] {
+            let result = static_files::handle(&HandleOpts {
+                method: &method,
+                headers: &headers,
+                base_path: &root_dir,
+                uri_path: "foo.html",
+                uri_query: None,
+                #[cfg(feature = "experimental")]
+                memory_cache: None,
+                #[cfg(feature = "directory-listing")]
+                dir_listing: false,
+                #[cfg(feature = "directory-listing")]
+                dir_listing_order: 6,
+                #[cfg(feature = "directory-listing")]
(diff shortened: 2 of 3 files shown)


mightyiam:
The test in the first commit fails before the second commit is applied.

tests/fixtures/.hidden-root/foo.html
joseluisq, line 1: For completion. Let's also add a simple `.dot` file in the `tests/fixtures/.hidden-root` directory and write a unit test for it.

tests/static_files.rs
```diff
@@ -1519,6 +1519,39 @@ mod tests {
...
+                #[cfg(feature = "directory-listing-download")]
+                dir_listing_download: &[],
+                redirect_trailing_slash: true,
+                compression_static: true,
+                ignore_hidden_files: true,
+                disable_symlinks: false,
+                index_files: &[],
+            })
+            .await
+            .expect("hidden base path not ignored");
```
joseluisq: Let's also assert the response, status, header, and body here. You can guide yourself from previous tests.
Do the same for the `.dot` test case.

joseluisq (changes requested):

joseluisq:
Good catch.
The title and description were not clear, so I just updated them for users can get more context.

joseluisq (approved):
Thanks!

PR #622, opened 2026-02-05
Add internal priority support of dynamic compression encodings (#618)

### LINKED ISSUE #618: Prioritize compression algorithms

### Feature scope

Right now it takes the order of compression algorithms from the order the browser sends it. However most browsers seem to send gzip as first option even if they support other (more modern/better compressions) as well. This leads to sub-optimal asset delivery.

Ether have a fixed order (or potentially even allow users to overwrite the order). The code then should go through the custom order and if that option is supported within the Accept-Encoding it should be served to the user (ether by runtime compression or if static compression enabled the appropriate static file if it exists).

Using alternatives like nginx for my asset delivery.

### Build target

All targets

### Additional context

Investigation provided by @akornatskyy

_In practice, many servers ignore ordering and q-values and instead use a fixed internal preference order._

Given:

```
Accept-Encoding: gzip, deflate, br, zstd
```

Most servers interpret this as:

> "Client supports all of these - I’ll choose my favorite"

**Nginx**

Default behavior (without special modules):

Priority:

    br (if ngx_brotli module enabled)
    gzip

Notes:

    Ignores order in Accept-Encoding
    Ignores q-values
    zstd requires a third-party module
    deflate is rarely used

**Apache HTTP Server**

Priority (if modules enabled):

    br (mod_brotli)
    gzip (mod_deflate)

Notes:

    Uses module preference, not header order
    zstd supported only in newer versions with mod_brotli-like extensions or patches

**Caddy**

Priority:

    zstd
    br
    gzip

**H2O**

Priority:

    br is preferred over gzip even if listed later

Header order is ignored.
Envoy

Priority (configurable, but defaults often):

    zstd
    br
    gzip

**Cloudflare / major CDNs**

Typical priority:

    zstd (if client supports it)
    br
    gzip

Notes:

    CDNs optimize for bandwidth + CPU
    Ignore header order entirely

### Browser behavior

Modern browsers usually send something like:
Chrome / Edge / Firefox

Accept-Encoding: gzip, deflate, br, zstd

But:

    Browser expects servers to choose the best
    Order does not indicate preference
    Browsers are fine with any supported encoding

msuarezd:
prioritizes modern compression algorithms while honoring client preferences

## Description
This will sort the accepted endcodings using both the client qualities and an intrernal priority.
It will prefer modern algorithms. 

## Related Issue
Closes #618

## Motivation and Context
Currenly the server will choose the compresion algorithm based solely on the qualities provided by the client.
Most browsers will not provide qualities, but just a list like `gzip, deflate, br, zstd`. 
In that case, the current sorting will not change the order and gzip will also be choosen, discarding more modern algorithms.

The internal priority as for now is:
- zstd (6)
- brotli (5)
- gzip (4)
- deflate (3)
- compress (2)
- none (1)

Given equal quality specified by the client, the encoding with the highest priorty will be put first in the sorted list and (if supported) will be used to compress the response.

## How Has This Been Tested?
Tested with curl and several combinations of accept-encoding

### DIFF
--- a/docs/content/features/compression.md
+++ b/docs/content/features/compression.md
@@ -15,6 +15,12 @@ static-web-server \
 
 The compression algorithm is determined by the [`Accept-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header and the compression support built into SWS. By default SWS builds with support for `Gzip`, `Deflate`, `Brotli` and `Zstandard` algorithms.
 
+SWS honors the qualities specified by the client to choose the algorithm. In case of equal quality for several algorithms (or no qualities at all), the internal priority will select according to this priority list:
+1. `Zstandard`
+2. `Brotli`
+3. `Gzip`
+4. `Deflate`
+
 ## MIME types compressed
 
 Compression is only applied to files with the MIME types listed below, indicating text and similarly well compressing formats. The asterisk `*` is a placeholder indicating an arbitrary MIME type part.
--- a/docs/content/features/compression.md
+++ b/docs/content/features/compression.md
@@ -15,7 +15,8 @@ static-web-server \
 
 The compression algorithm is determined by the [`Accept-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header and the compression support built into SWS. By default SWS builds with support for `Gzip`, `Deflate`, `Brotli` and `Zstandard` algorithms.
 
-SWS honors the qualities specified by the client to choose the algorithm. In case of equal quality for several algorithms (or no qualities at all), the internal priority will select according to this priority list:
+SWS honors the qualities specified by the client to choose the algorithm. In case of equal quality for several algorithms (or no qualities at all), the internal priority will be selected according to this list:
+
 1. `Zstandard`
 2. `Brotli`
 3. `Gzip`
--- a/src/headers_ext/content_coding.rs
+++ b/src/headers_ext/content_coding.rs
@@ -81,6 +81,20 @@ macro_rules! define_content_coding {
     }
 }
 
+impl ContentCoding {
+    pub fn priority(&self) -> u8 {
+        match self {
+            Self::IDENTITY => 1,
+            Self::COMPRESS => 2,
+            Self::DEFLATE => 3,
+            Self::GZIP => 4,
+            Self::BROTLI => 5,
+            Self::ZSTD => 6,
+            _ => 0,
+        }
+    }
+}
+
 define_content_coding! {
     BROTLI; "br",
     COMPRESS; "compress",
--- a/src/headers_ext/accept_encoding.rs
+++ b/src/headers_ext/accept_encoding.rs
@@ -64,8 +64,8 @@ mod tests {
         let accept_enc = AcceptEncoding(val.into());
 
         let mut encodings = accept_enc.sorted_encodings();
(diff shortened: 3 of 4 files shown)


src/headers_ext/content_coding.rs
```diff
@@ -81,6 +81,20 @@ macro_rules! define_content_coding {
...
+            Self::COMPRESS => 2,
+            Self::DEFLATE => 3,
+            Self::GZIP => 4,
+            Self::BROTLI => 5,
+            Self::ZSTD => 6,
+            _ => 0,
+        }
+    }
+}
+
```
joseluisq, line 97: I like the list's order.

But can you please just update the PR description to show the internal priority list?

src/headers_ext/content_coding.rs
```diff
@@ -81,6 +81,20 @@ macro_rules! define_content_coding {
...
+            Self::IDENTITY => 1,
+            Self::COMPRESS => 2,
+            Self::DEFLATE => 3,
+            Self::GZIP => 4,
+            Self::BROTLI => 5,
+            Self::ZSTD => 6,
+            _ => 0,
+        }
+    }
+}
```
joseluisq, line 96: Also, the new internal priority order has to be documented.

Please use the _compression_ page to **briefly** describe the internal priority ordering.
E.g. A new entry below the `"Choice of compression algorithm"`.

https://github.com/static-web-server/static-web-server/blob/master/docs/content/features/compression.md

joseluisq (changes requested):
I like the priority approach, and it still respects user preferences when needed.
I just left a few comments for completeness to finish the PR.

src/headers_ext/content_coding.rs
msuarezd: Sure, done.

src/headers_ext/content_coding.rs
msuarezd: Added. Is it understandable?

docs/content/features/compression.md
```diff
@@ -15,6 +15,12 @@ static-web-server \
 
 The compression algorithm is determined by the [`Accept-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header and the compression support built into SWS. By default SWS builds with support for `Gzip`, `Deflate`, `Brotli` and `Zstandard` algorithms.
 
+SWS honors the qualities specified by the client to choose the algorithm. In case of equal quality for several algorithms (or no qualities at all), the internal priority will select according to this priority list:
+1. `Zstandard`
+2. `Brotli`
+3. `Gzip`
+4. `Deflate`
+
```
joseluisq: It looks fine, I just fixed a small typo and added an empty line.

```suggestion
SWS honors the qualities specified by the client to choose the algorithm. In case of equal quality for several algorithms (or no qualities at all), the internal priority will be selected according to this list:

1. `Zstandard`
2. `Brotli`
3. `Gzip`
4. `Deflate`

```

We have a Markdown formatter https://github.com/hukkin/mdformat. In case the checks fail, you can use that tool to format the `.md` file.

joseluisq (changes requested):

src/headers_ext/content_coding.rs
joseluisq: Yes, I did some small improvements, so we should be ready to merge.

joseluisq (approved):
All fine, ¡gracias!

PR #624, opened 2026-02-10
Separate static pre-compression from dynamic compression features

### LINKED ISSUE #362: Static compression should not be dependent on the compression features

msuarezd:
This enables SWS to deliver static precompressed files even if no compression features are enabled.

## Description
It separates compression features from static compression, as there are no compression algorithms needed to deliver already compressed files.

## Related Issue
This closes #362

## How Has This Been Tested?
Built and tested with compression features enabled and disabled. In both cases the server is able to deliver precompressed files.

### DIFF
--- a/src/settings/cli.rs
+++ b/src/settings/cli.rs
@@ -296,23 +296,6 @@ pub struct General {
     /// Compression level to apply for Gzip, Deflate, Brotli or Zstd compression.
     pub compression_level: super::CompressionLevel,
 
-    #[cfg(any(
-        feature = "compression",
-        feature = "compression-gzip",
-        feature = "compression-brotli",
-        feature = "compression-zstd",
-        feature = "compression-deflate"
-    ))]
-    #[cfg_attr(
-        docsrs,
-        doc(cfg(any(
-            feature = "compression",
-            feature = "compression-gzip",
-            feature = "compression-brotli",
-            feature = "compression-zstd",
-            feature = "compression-deflate"
-        )))
-    )]
     #[arg(
         long,
         default_value = "false",
--- a/src/static_files.rs
+++ b/src/static_files.rs
@@ -390,14 +390,10 @@ fn get_composed_file_metadata<'a>(
                 }
             }
 
-            let precompressed_variant = if compression_static {
-                match compression_static::precompressed_variant(file_path, headers) {
-                    Some(p) => Some((p.file_path, p.encoding)),
-                    None => None,
-                }
-            } else {
-                None
-            };
+            let precompressed_variant = compression_static
+                .then(|| compression_static::precompressed_variant(file_path, headers))
+                .flatten()
+                .map(|p| (p.file_path, p.encoding));
 
             Ok(FileMetadata {
                 file_path,
--- a/src/static_files.rs
+++ b/src/static_files.rs
@@ -25,13 +25,6 @@ use crate::response::response_body;
 #[cfg(feature = "experimental")]
 use crate::mem_cache::{cache, cache::MemCacheOpts};
 
-#[cfg(any(
-    feature = "compression",
-    feature = "compression-deflate",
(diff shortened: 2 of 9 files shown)


msuarezd:
@joseluisq what do you think about this?

joseluisq:
> @joseluisq what do you think about this?

It looks fine to me after a quick look, but I will give it a proper check in the next days.

As an idea, probably once this gets merged, we can think about to add a new `compression-static` Cargo feature (another PR).

src/settings/cli.rs
joseluisq, line 1: Due to the changes, the Cargo `compression` feature wording needs to be updated. 

https://github.com/static-web-server/static-web-server/blob/735cc79128d4a2f8ba6dcd0a5730dbc81be7bdd7/docs/content/building-from-source.md?plain=1#L39-L43

src/static_files.rs
```diff
@@ -403,46 +388,28 @@ fn get_composed_file_metadata<'a>(
...
 
+            let precompressed_variant = if compression_static {
+                match compression_static::precompressed_variant(file_path, headers) {
+                    Some(p) => Some((p.file_path, p.encoding)),
+                    None => None,
+                }
+            } else {
+                None
+            };
+
```
joseluisq:
```suggestion
            let precompressed_variant = compression_static
                .then(|| compression_static::precompressed_variant(file_path, headers))
                .flatten()
                .map(|p| (p.file_path, p.encoding));

```

joseluisq (changes requested):
It looks fine, only a few minor changes to make.

src/settings/cli.rs
joseluisq: The same for the Crate docs.

https://github.com/static-web-server/static-web-server/blob/735cc79128d4a2f8ba6dcd0a5730dbc81be7bdd7/src/lib.rs#L84-L88

msuarezd: Updated.

msuarezd:
> > @joseluisq what do you think about this?
> 
> It looks fine to me after a quick look, but I will give it a proper check in the next days.
> 
> As an idea, probably once this gets merged, we can think about to add a new `compression-static` Cargo feature (another PR).

I can take care of that. I think however ist is not really necessary, as static compression does not add any build dependency, is only a run time feature which already can be turned on an off with startup parameters, don't you think?

joseluisq:
> I can take care of that. I think however ist is not really necessary, as static compression does not add any build dependency, is only a run time feature which already can be turned on an off with startup parameters, don't you think?

Yes, that's right. Let's keep it simple. There's no need for a new feature in that case.

joseluisq:
For some reason, the FreeBSD pipelines got stuck when installing the Rust toolchain. Probably network, I will investigate that.

joseluisq (approved):
Ready for the release.

Thanks!

PR #637, opened 2026-02-27
Configurable access log formatting

chrissnell:
## Problem

SWS uses a bespoke logging format.  Many users (particularly industry users) want a standard access log format like nginx style, or maybe NCSA.  The default logging outputs request data only (method, URI, remote address) and uses a fixed
format. Users deploying SWS behind reverse proxies or feeding logs into aggregation pipelines
need control over the log format and access to response-level fields like status codes, response
size, and request duration.

## Change Summary

Add configurable access log format support with a `%{token}` syntax, allowing users to define
custom log output that includes both request and response data (status, bytes, duration).

When `access-log-format` is set, the server switches from the default pre-request logging to a
post-response log line rendered from the compiled format string. When unset, behavior is unchanged.

## Changes

- New `access_logs` module replacing the former `log_addr` module, consolidating request logging
  and trusted proxy IP extraction into a single module
- Format string parser (`AccessLogFormat::parse`) that compiles a `%{token}` format at startup
  into an efficient segment list for zero-allocation rendering at request time
- 13 supported tokens: `method`, `uri`, `status`, `bytes`, `duration`, `remote_addr`,
  `x_real_ip`, `forwarded_for`, `host`, `user_agent`, `referer`, `timestamp`, `version`
  (Note for @joseluisq -- your original code consolidates `x_real_ip` and `forwarded_for`; this distinguishes between the two)
- `timestamp` token uses local time, honoring the `TZ` environment variable introduced in your recent PR
- CLI flag (`--access-log-format`), environment variable (`SERVER_ACCESS_LOG_FORMAT`), and
  TOML config (`access-log-format`) support
- Comprehensive unit tests for parsing, rendering, and IP extraction

## Example

```toml
access-log-format = '%{remote_addr} - [%{timestamp}] "%{method} %{uri} %{version}" %{status} %{bytes} "%{referer}" "%{user_agent}"'
```
## Considerations
- I left the current logging format as the default.  I don't like it (it's very non-standard) but you and your users are probably used to it and may have logging filters built for it already
- The current logging also emits logs from `static_web_server::virtual_hosts`.  A lot of this info is duplicated by the `static_web_server::access_logs` logs.  Is there a reason to maintain this logging?

## To-do
- Create an example filter in VRL for filtering these logs in [Vector](https://vector.dev/)


chrissnell:
@joseluisq I rebased this and fixed the formatting but it looks like GH actions are still crashing for no apparent reason..  🤷

src/access_logs.rs
joseluisq, line 1: Honestly, I do not like the access_logs implementation. It seems a bit convoluted.
I believe we can improve it and simplify it more. Also, consider security when it comes to the parser.

Take your time and let me know if there's any concern.

joseluisq:
> @joseluisq I rebased this and fixed the formatting but it looks like GH actions are still crashing for no apparent reason.. 🤷

The CI can not test the changes without the default features. See https://github.com/static-web-server/static-web-server/actions/runs/22505118629/job/65201741063?pr=637#step:14:238

Try to test the changes without the default feature enabled, something like `cargo test --verbose --no-default-features`.

src/access_logs.rs
chrissnell: > Honestly, I do not like the access_logs implementation. It seems a bit convoluted. I believe we can improve it and simplify it more. Also, consider security when it comes to the parser.
> 
> Take your time and let me know if there's any concern.

For the token parser (i.e., the configuration of the tokens), are you okay with the `parse()` method or did you want something similar like using regex or `str::split`?

Regarding security concerns, are you referring to the ability of a user to inject fake log messages via user-controlled elements like `user_agent`, `referrer`, etc?  If so, I could sanitize these values and strip control characters and newlines.

joseluisq:
> * I left the current logging format as the default.  I don't like it (it's very non-standard) but you and your users are probably used to it and may have logging filters built for it already

Yes, the current log format could be deprecated and replaced by this one in the near future.

> * The current logging also emits logs from `static_web_server::virtual_hosts`.  A lot of this info is duplicated by the `static_web_server::access_logs` logs.  Is there a reason to maintain this logging?

Do you mean this single entry? https://github.com/static-web-server/static-web-server/blob/master/src/virtual_hosts.rs#L47
The question is, does really access_logs cover virtual host logs too?

joseluisq:
> * Create an example filter in VRL for filtering these logs in [Vector](https://vector.dev/)

Sure, you can add additional stuff in a separate PR.

chrissnell:
> > * The current logging also emits logs from `static_web_server::virtual_hosts`.  A lot of this info is duplicated by the `static_web_server::access_logs` logs.  Is there a reason to maintain this logging?
> 
> Do you mean this single entry? https://github.com/static-web-server/static-web-server/blob/master/src/virtual_hosts.rs#L47 The question is, does really access_logs cover virtual host logs too?

Yes sir, that was the one.  Out of scope for this PR but perhaps the default mode could be to log all access logs to stdout and all informational logs (such as that one) to stderr.

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+                            "access log format: unexpected character '{}' after '%' at position {}; use %{{name}} for tokens or %%%% for literal %%",
+                            bytes[i + 1] as char,
+                            i
+                        );
+                    }
+                }
+            } else {
+                literal.push(bytes[i] as char);
+                i += 1;
+            }
```
joseluisq, line 122: Here is a small improvement for the parser.

```suggestion
    pub(crate) fn parse(format: &str) -> crate::Result<Self> {
        let mut segments = Vec::new();
        let mut chars = format.chars().peekable();
        let mut literal = String::new();

        while let Some(c) = chars.next() {
            if c == '%' {
                match chars.peek() {
                    Some('%') => {
                        literal.push('%');
                        chars.next();
                    }
                    Some('{') => {
                        chars.next(); // skip '{'
                        // Flush accumulated literal
                        if !literal.is_empty() {
                            segments.push(Segment::Literal(std::mem::take(&mut literal)));
                        }
                        let mut name = String::new();
                        while let Some(&ch) = chars.peek() {
                            if ch == '}' {
                                chars.next();
                                break;
                            }
                            name.push(ch);
                            chars.next();
                        }
                        if name.is_empty() || !format.contains(&format!("%{{{name}}}")) {
                            bail!("access log format: unclosed '%{{' or empty token");
                        }
                        let token = Token::from_name(&name).ok_or_else(|| {
                            anyhow::anyhow!("access log format: unknown token '%{{{name}}}'")
                        })?;
                        segments.push(Segment::Token(token));
                    }
                    Some(&other) => {
                        bail!(
                            "access log format: unexpected character '{}' after '%' at position; use %{{name}} for tokens or %% for literal %",
                            other
                        );
                    }
                    None => {
                        bail!("access log format: trailing '%' at end of string");
                    }
                }
            } else {
                literal.push(c);
            }
        }
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+        hyper::Version::HTTP_10 => "HTTP/1.0",
+        hyper::Version::HTTP_11 => "HTTP/1.1",
+        hyper::Version::HTTP_2 => "HTTP/2",
+        hyper::Version::HTTP_3 => "HTTP/3",
+        _ => "HTTP/?",
+    }
+}
+
+// --- Trusted proxy helpers (absorbed from log_addr.rs) ---
+
```
joseluisq, line 238:
```suggestion
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+        .and_then(|h| h.to_str().ok())
+        .and_then(|s| s.split(',').next())
+        .and_then(|s| s.trim().parse::<IpAddr>().ok())
+}
+
+fn header_str(req: &Request<Body>, name: hyper::header::HeaderName) -> &str {
+    req.headers()
+        .get(name)
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("")
```
joseluisq, line 277:
```suggestion
        .unwrap_or_default()
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+}
+
+// --- Public API ---
+
```
joseluisq, line 281:
```suggestion
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+) {
+    // Skip custom format logging for health endpoints at debug level
+    let is_health = opts.health && health::is_health_endpoint(req);
+
+    let bytes = resp
+        .headers()
+        .get(hyper::header::CONTENT_LENGTH)
+        .and_then(|v| v.to_str().ok())
+        .and_then(|v| v.parse::<u64>().ok())
+        .unwrap_or(0);
```
joseluisq, line 364:
```suggestion
        .unwrap_or_default();
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+        if content_length > 0 {
+            resp.headers_mut().insert(
+                hyper::header::CONTENT_LENGTH,
+                content_length.to_string().parse().unwrap(),
+            );
+        }
+        resp
+    }
+
+    // --- Parsing tests ---
```
joseluisq, line 430:
```suggestion
    // Parsing tests
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn parse_bare_percent_char_fails() {
+        let result = AccessLogFormat::parse("%x");
+        assert!(result.is_err());
+    }
+
+    // --- Rendering tests ---
```
joseluisq, line 514:
```suggestion
    // Rendering tests
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+            host: "",
+            user_agent: "",
+            referer: "",
+            x_real_ip: None,
+            forwarded_for: None,
+        };
+        assert_eq!(fmt.render(&ctx), "100% GET");
+    }
+
+    // --- IP extraction tests ---
```
joseluisq, line 588:
```suggestion
    // IP extraction tests
```

src/access_logs.rs
```diff
@@ -0,0 +1,650 @@
...
+    fn extract_disabled_returns_none() {
+        let req = make_request_with_headers("GET", "/", vec![("X-Real-IP", "10.0.0.1")]);
+        let opts = RequestHandlerOpts {
+            log_x_real_ip: false,
+            ..Default::default()
+        };
+        assert_eq!(extract_x_real_ip(&req, &opts, None), None);
+    }
+
+    // --- post_process integration test ---
```
joseluisq, line 639:
```suggestion
    // post_process integration test
```

src/access_logs.rs
joseluisq: > For the token parser (i.e., the configuration of the tokens), are you okay with the `parse()` method or did you want something similar like using regex or `str::split`?

I shared an improvement to make. See https://github.com/static-web-server/static-web-server/pull/637/changes#r2875191980

joseluisq: > Regarding security concerns, are you referring to the ability of a user to inject fake log messages via user-controlled elements like `user_agent`, `referrer`, etc? If so, I could sanitize these values and strip control characters and newlines.

Yeah, if you can sanitize those values where it makes sense would be great.

src/settings/cli.rs
joseluisq, line 1: We have to update the `logging.md` page explaining how this feature will work.

joseluisq (changes requested):
I'm still taking some time to test the feature. In the meantime, here are a few improvements to make.

joseluisq:
FYI, now that v3 has made an appearance and v2 has become LTS. Consider this feature to target the v3 (`master`) branch instead. Or close it and open a new one targeting `master`.

PR #645, opened 2026-04-17
Add FIPS-capable TLS via http2-fips feature flag

### LINKED ISSUE #644: Add FIPS-capable TLS support through a feature flag

### Feature scope

Security

SWS is really nicely positioned to be deployed in high-compliance environments (such as US federal or highly regulated industries) because of its small codebase, from-scratch container images, and, of course, being in Rust. The only thing missing is offering TLS using a FIPS-140-validated cryptographic module.

Fortunately, [`aws-lc-rs`](https://github.com/aws/aws-lc-rs) is API-compatible with `ring`, and [provides bindings](https://github.com/aws/aws-lc-rs?tab=readme-ov-file#aws-lc-fips-sys) to [AWS-LC-FIPS 3.x](https://github.com/aws/aws-lc/tree/fips-2024-09-27). It would be super useful for me (and hopefully others!) for SWS to provide an alternate build and container images using this as a feature-flagged alternate cryptographic backend.

So, the solution I spent some time playing with is to add an `http2-fips` Cargo feature flag which substitutes `aws-lc-rs` in FIPS mode for `ring` as the TLS crypto provider. Then, add FIPS build targets for a handful of platforms where `aws-lc-fips-sys` supports static linking, and corresponding container builds for those targets. The main hurdle is that the FIPS builds then require CMake and Go at build time, but the outputs are still statically compiled binaries.

Another potential option here is to replace `ring` project-wide with `aws-lc-rs` (without enabling FIPS mode!) `rustls` [did that in 0.23](https://github.com/rustls/rustls/releases/tag/v%2F0.23.0). That's something that requires more consideration than I wanted to tackle in my proof of concept.

### Build target

All targets

#### Prototype

I went and built out what seems to me to be a working prototype on [my branch](https://github.com/alexander-bauer/static-web-server/tree/feat/fips-crypto-provider). I'm a bit of a Rust newbie, so I hope that this is a reasonable approach:

- Restructure the existing `http2` feature into a base feature for TLS plumbing, plus provider-specific features `http2-ring` (default) and `http2-fips`.
- Add `compile_error!` guards to enforce that exactly one provider flag is set at once.
- Add a test when `http2-fips` is enabled which asserts `aws_lc_rs::try_fips_mode()`
- Add CI config:
  - FIPS build & test
  - Releases for FIPS binary builds on the four Linux targets where `aws-lc-fips-sys` supports static linking: `{x86_64,aarch64}-unknown-linux-{gnu,musl}`
  - Build FIPS Docker images for each of the Alpine, Debian, and scratch families for `linux/amd64` and `linux/arm64`, matching the binary builds

I think that it's nearly PR-ready if this is something you're interested in having; it would need some docs added, but I wanted to raise this for discussion before I do any polishing work.

#### Notes on FIPS-140-3

Usually when software engineers talk about FIPS, they mean FIPS-140-3. Hopefully not everybody has had to interact with FIPS a ton, because it can be a pain. There's a ton of nuance surrounding FIPS validation and certification which is way over my head, since I'm an engineer and not a compliance officer, but the very short practical overview is this: FIPS-140-3 is a software certification given out by NIST, a U.S. Federal standards body. It can apply to applications, libraries, and cryptographic providers, and different high-compliance environments have different requirements for how certified the software they use is.

The lowest hanging fruit here is for applications to use FIPS-validated cryptographic providers. That doesn't make the whole application validated, and "validation" is less complete than the full NIST _certification_, but it's sufficient for a lot of uses, including [FedRAMP](https://www.fedramp.gov/resources/documents/FedRAMP_Policy_for_Cryptographic_Module_Selection_v1.1.0.pdf).

If SWS adopted `aws-lc-rs` in FIPS mode, my understanding is that it would be suitable for use in FedRAMP environments, and probably a variety of other regulated environments. It's already a really nice fit for them because of the language, codebase size, and scratch images. With a FIPS-validated crypto backend, SWS could be a clear replacement for a lot of NGINX/IIS/Apache deployments which have a wider CVE surface area.

alexander-bauer:
## Description

Proof of concept for FIPS-capable TLS support. Adds an `http2-fips` feature flag that uses `aws-lc-rs` in FIPS mode instead of `ring` as the TLS crypto provider, along with CI to build and ship FIPS binaries and Docker images. Targets are limited to the Linux platforms where `aws-lc-fips-sys` supports static linking (x86_64 and aarch64, gnu and musl), preserving the single-binary / scratch container property.

## Related Issue

#644

## Motivation and Context

SWS is well-positioned for high-compliance environments (US federal, regulated industries) thanks to its small codebase, scratch container images, and being in Rust. The missing piece is TLS using a FIPS-validated cryptographic module -- `ring` doesn't have FIPS validation.

`aws-lc-rs` is API-compatible with `ring` and provides bindings to the [AWS-LC FIPS module](https://github.com/aws/aws-lc). This PR restructures the `http2` feature into a base plus provider-specific features (`http2-ring` as the default, `http2-fips` as opt-in), so existing builds are completely unaffected. More detail and context on the FIPS landscape in #644.

## How Has This Been Tested?

- `cargo test --features=all` passes all 32 existing tests (default ring provider, unchanged behavior)
- `cargo test --no-default-features --features http2-fips,...` passes all 32 tests plus a new `fips_mode_is_active` test that asserts `aws_lc_rs::try_fips_mode()` returns `Ok`
- `cargo check --features http2-ring,http2-fips` correctly triggers `compile_error!` for conflicting providers
- `cargo check --no-default-features --features http2` correctly triggers `compile_error!` for missing provider
- `cargo check --no-default-features` compiles successfully (HTTP/1 only, no TLS)

## Some Other Notes

- The `github-advanced-security` bot raised an error on the mutually-exclusive feature flags. I'm not sure how we can avoid that, or if it's going to raise that error on every PR.

### DIFF
--- a/.github/workflows/devel.fips.yml
+++ b/.github/workflows/devel.fips.yml
@@ -1,0 +1,46 @@
+name: devel-fips
+
+on:
+  pull_request:
+    paths:
+      - "**.rs"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "Cross.fips.toml"
+      - ".github/workflows/devel.fips.yml"
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    name: FIPS build check
+    runs-on: ubuntu-22.04
+    env:
+      CARGO_FEATURES: "--no-default-features --features http2-fips,compression,directory-listing,directory-listing-download,basic-auth,fallback-page,metrics,experimental"
+      RUSTFLAGS: "--cfg tokio_unstable"
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@stable
+
+    - name: Install FIPS build dependencies
+      run: sudo apt-get update && sudo apt-get install -y cmake golang
+
+    - name: Run tests
+      run: cargo test --verbose ${{ env.CARGO_FEATURES }}
+
+    - name: Build release binary
+      run: cargo build --bin static-web-server --release ${{ env.CARGO_FEATURES }}
+
+    - name: Verify binary is statically linked (no aws-lc shared objects)
+      run: |
+        echo "Checking linked libraries..."
+        ldd target/release/static-web-server
+        if ldd target/release/static-web-server | grep -q 'aws_lc'; then
+          echo "ERROR: aws-lc shared library detected -- FIPS module is not statically linked"
+          exit 1
+        fi
+        echo "OK: no aws-lc shared objects found"
--- a/.github/workflows/devel.fips.yml
+++ b/.github/workflows/devel.fips.yml
@@ -3,14 +3,22 @@ name: devel-fips
 on:
   pull_request:
     paths:
-      - "**.rs"
-      - "Cargo.toml"
-      - "Cargo.lock"
+      - Cargo.lock
+      - Cargo.toml
+      - src/**
+      - tests/**
       - "Cross.fips.toml"
       - ".github/workflows/devel.fips.yml"
   push:
     branches:
       - master
+    paths:
+      - Cargo.lock
+      - Cargo.toml
+      - src/**
+      - tests/**
+      - "Cross.fips.toml"
+      - ".github/workflows/devel.fips.yml"
 
 jobs:
   test:
--- a/.github/workflows/devel.fips.yml
+++ b/.github/workflows/devel.fips.yml
@@ -56,6 +56,9 @@ jobs:
     - name: Build release binary
       run: cargo build --bin static-web-server --release ${{ env.CARGO_FEATURES }}
 
+    - name: Verify binary has FIPS mode enabled
+      run: target/release/static-web-server -V | grep -i "fips"
+
     - name: Verify binary is statically linked (no aws-lc shared objects)
       run: |
         echo "Checking linked libraries..."
--- a/.github/workflows/devel.fips.yml
+++ b/.github/workflows/devel.fips.yml
@@ -17,7 +17,7 @@ jobs:
     name: FIPS build check
     runs-on: ubuntu-22.04
     env:
-      CARGO_FEATURES: "--no-default-features --features http2-fips,compression,directory-listing,directory-listing-download,basic-auth,fallback-page,metrics,experimental"
+      CARGO_FEATURES: "--no-default-features --features all-fips"
       RUSTFLAGS: "--cfg tokio_unstable"
     steps:
     - name: Checkout repository
@@ -62,7 +62,7 @@ jobs:
     name: FIPS rustdoc check
     runs-on: ubuntu-22.04
     env:
-      CARGO_FEATURES: "--no-default-features --features http2-fips,compression,directory-listing,directory-listing-download,basic-auth,fallback-page,metrics,experimental"
+      CARGO_FEATURES: "--no-default-features --features all-fips"
       RUSTFLAGS: "--cfg tokio_unstable"
     steps:
     - name: Checkout repository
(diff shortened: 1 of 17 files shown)


alexander-bauer:
@joseluisq Okay! I think that this is reasonably ready for review. I'm still not sure whether the GitHub security bot is going to be an issue, and I'm not much of a GitHub Actions expert. I hope it's a reasonable adaptation of the existing workflows.

.github/workflows/devel.fips.yml
```diff
@@ -0,0 +1,89 @@
+name: devel-fips
+
+on:
+  pull_request:
+    paths:
+      - "**.rs"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "Cross.fips.toml"
+      - ".github/workflows/devel.fips.yml"
```
joseluisq, line 11:
```suggestion
    paths:
      - Cargo.lock
      - Cargo.toml
      - src/**
      - tests/**
      - "Cross.fips.toml"
      - ".github/workflows/devel.fips.yml"
```

.github/workflows/devel.fips.yml
```diff
@@ -0,0 +1,89 @@
...
+  push:
+    branches:
+      - master
```
joseluisq, line 14: For some reason, this new workflow isn't triggered. Let's try this:

```suggestion
  push:
    branches:
      - master
    paths:
      - Cargo.lock
      - Cargo.toml
      - src/**
      - tests/**
      - "Cross.fips.toml"
      - ".github/workflows/devel.fips.yml"
```

docker/alpine-fips/Dockerfile
joseluisq, line 1: The Docker release workflow looks fine. But we are not building the new FIPS Docker images on CI.
Let's add a new workflow similar to [this](https://github.com/static-web-server/static-web-server/blob/master/.github/workflows/devel.docker.yml), for the FIPS Docker images.

Of course, there won't be any binaries to fetch from this time, but I will check them out locally and after a release anyway.
Btw, if you can, check out Github Actions `fail-fast: false` or `continue-on-error: true` options, maybe they could help to bypass errors for now. If not, all fine, I will manage it later.

joseluisq (changes requested):
> @joseluisq Okay! I think that this is reasonably ready for review. I'm still not sure whether the GitHub security bot is going to be an issue, and I'm not much of a GitHub Actions expert. I hope it's a reasonable adaptation of the existing workflows.

@alexander-bauer I gave it a first look. In general is fine, just a few minor improvements or missing details about CI workflows.

Please ignore the CI doc's error checks.

I will take a look at the cargo and Docker builds locally.

alexander-bauer:
@joseluisq I added a `devel.docker.fips.yml` workflow to mirror `devel.docker.yml` with `continue-on-error: true` for now so that it can succeed until there's a release artifact for it to download. I think that's what you meant, but please let me know if not! :smile:

joseluisq (commented):
@alexander-bauer FYI, I made a small improvement so that users can quickly verify the SWS binary for FIPS and added an alternative build with Clang to the docs.

I took the info from https://github.com/aws/aws-lc-rs/blob/main/aws-lc-fips-sys/README.md.

It looks like [this](https://github.com/static-web-server/static-web-server/actions/runs/25889982800/job/76090782491#step:9:1).

joseluisq (approved):
It looks fine from my side.

We could also update the docs about Docker by mentioning the new _fips_ images, but this can be in another PR. Feel free if you want.

If there is some work to do on CI, I will manage it later.

Also, looking forward to adopting `aws-lc-rs` instead of ring (non-FIPS) by default in the future if it can build on all SWS target platforms.

Thanks!

alexander-bauer:
That's great, thank you so much!

PR #656, opened 2026-05-11
Optional default charset for text/* file responses (#655)

### LINKED ISSUE #655: Optional default charset for `text/*` responses

### Feature scope

Configuration (e.g. TOML) or CLI/env option

SWS follows RFC 6657 and never appends a `charset=` parameter to `text/*` Content-Type responses. For text formats that have no in-band encoding declaration — `robots.txt`, `llms.txt`, sitemaps emitted as `.txt`, plain READMEs — any non-ASCII byte ends up rendered as mojibake in browsers, which default to Windows-1252 / Latin-1 when no `charset` header is present. A correctly-encoded UTF-8 file containing `—` (`E2 80 94`) is displayed as `â€"` despite the bytes on disk being valid. The only mitigation today is per-glob `[[advanced.headers]]` entries, which are verbose, easy to forget, and have to be re-pasted into every project.

A global "declared default charset for `text/*`" option, parallel to Apache's `AddDefaultCharset utf-8` and nginx's `charset utf-8;` + `charset_types`.

imho, this should be enabled by default but I let you decide.

Use `[[advanced.headers]]` for `**/*.md` and `**/*.txt` — works but is verbose, no env vars, etc. And easy to forget for an option which is needed to render text/markdown right in browsers.

### Build target

All targets

### Additional context

_No response_

davlgd:
## Description

This PR adds an optional `--text-charset` setting that declares a default `charset` parameter on `text/*` responses which don't already carry one. The option is enabled by default with `utf-8` and can be set to any RFC 7230 token (e.g. `iso-8859-1`) or left empty to disable the feature. It is also exposed as the `SERVER_TEXT_CHARSET` env variable and the `text-charset` key under `[general]` in the TOML configuration file.

The behaviour is conservative:

- Only `text/*` responses are touched. `application/json`, JavaScript, images and other types are never modified.
- Responses that already carry a `charset` parameter (markdown content negotiation, error pages, maintenance mode, etc.) are left untouched.
- `[[advanced.headers]]` overrides still win, since they are applied after this step.
- Invalid charset values are rejected at startup with a clear error.

## Related Issue

Closes #655

## Motivation and Context

Files like `robots.txt`, `llms.txt`, sitemaps emitted as `.txt`, plain READMEs, and markdown sources are served with a bare `text/plain` or `text/markdown` `Content-Type` and no charset parameter. Browsers fall back to a locale-specific guess (typically Windows-1252) and render any non-ASCII byte as mojibake — a UTF-8 em-dash `—` (`E2 80 94`) shows up as `â€"`. The only workaround today is to repeat `[[advanced.headers]]` entries in every project, which is verbose and easy to forget.

This PR mirrors the spirit of Apache's `AddDefaultCharset` and nginx's `charset` directive, but applies the annotation to every `text/*` subtype (Apache only covers `text/plain` and `text/html`, nginx requires an explicit `charset_types` list that excludes `text/markdown` and `text/css` by default). Defaulting to `utf-8` fixes the issue out of the box for the vast majority of modern static sites, while still allowing opt-out via an empty value for operators serving legacy encodings. 

Currently, I have to add a `sws.toml` to each of my project with, at least: 

```
[advanced]

[[advanced.headers]]
source = "**/*.txt"
headers = { Content-Type = "text/plain; charset=utf-8" }

[[advanced.headers]]
source = "**/*.md"
headers = { Content-Type = "text/markdown; charset=utf-8" }
```

## How Has This Been Tested?

- **Unit tests** in `src/text_charset.rs` cover the algorithm: annotation of bare `text/*`, preservation of existing `charset` (including unusual casing), non-text types ignored, disabled mode, missing `Content-Type`, unrelated parameters present, and tchar validation.
- **Integration tests** in `tests/text_charset.rs` exercise the wiring through the request pipeline: default behaviour, disabled via empty string, custom charset value, `[[advanced.headers]]` taking precedence, and startup-time rejection of invalid values.
- Existing assertions in `tests/compression.rs`, `tests/cors.rs`, `tests/handler.rs`, `tests/markdown.rs`, `tests/redirects.rs`, `tests/rewrites.rs` and `tests/compression_static.rs` were updated to reflect the new default.
- Full suite: `cargo test --all-features` is green; `cargo clippy --all-features --tests --lib --bins -- -D warnings` and `cargo fmt --check` are clean.

### DIFF
--- a/src/text_charset.rs
+++ b/src/text_charset.rs
@@ -1,0 +1,173 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+// This file is part of Static Web Server.
+// See https://static-web-server.net/ for more information
+// Copyright (C) 2019-present Jose Quintana <joseluisq.net>
+
+//! Declares a default `charset` parameter on `text/*` responses that don't
+//! already have one. Similar in spirit to Apache's `AddDefaultCharset` and
+//! nginx's `charset` directives, applied here to every `text/*` subtype.
+
+use hyper::{
+    Body, Response,
+    header::{CONTENT_TYPE, HeaderValue},
+};
+
+use crate::{Error, handler::RequestHandlerOpts};
+
+/// Validates a charset value against the RFC 7230 `token` grammar.
+pub(crate) fn is_valid_charset(value: &str) -> bool {
+    !value.is_empty()
+        && value.bytes().all(|b| {
+            b.is_ascii_alphanumeric()
+                || matches!(
+                    b,
+                    b'!' | b'#'
+                        | b'$'
+                        | b'%'
+                        | b'&'
+                        | b'\''
+                        | b'*'
+                        | b'+'
+                        | b'-'
+                        | b'.'
+                        | b'^'
+                        | b'_'
+                        | b'`'
+                        | b'|'
+                        | b'~'
+                )
+        })
+}
+
+/// Adds `; charset=<value>` to the `Content-Type` header when the response
+/// type is `text/*` and no `charset` parameter is already present. A no-op
+/// when the option is disabled (empty value).
+pub(crate) fn post_process(
+    opts: &RequestHandlerOpts,
+    mut resp: Response<Body>,
+) -> Result<Response<Body>, Error> {
+    if opts.text_charset.is_empty() {
+        return Ok(resp);
+    }
+
+    let ct = match resp
+        .headers()
+        .get(CONTENT_TYPE)
+        .and_then(|v| v.to_str().ok())
+    {
+        Some(v) if is_text(v) && !has_charset_param(v) => v,
+        _ => return Ok(resp),
+    };
+
+    if let Ok(value) = HeaderValue::from_str(&format!("{ct}; charset={}", opts.text_charset)) {
+        resp.headers_mut().insert(CONTENT_TYPE, value);
+    }
+    Ok(resp)
+}
+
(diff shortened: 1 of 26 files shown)


davlgd:
I've added a fix in changelog, as CI was not happy with a typo from master.

src/text_charset.rs
```diff
@@ -0,0 +1,173 @@
...
+        resp.headers_mut().insert(CONTENT_TYPE, value);
+    }
+    Ok(resp)
+}
+
+fn is_text(content_type: &str) -> bool {
+    content_type
+        .get(..5)
+        .is_some_and(|p| p.eq_ignore_ascii_case("text/"))
+}
```
joseluisq, line 72: FYI, I would use `mime` instead here, as it is more resilient. But I will take care of it, since the compression module can also be improved on that direction.

joseluisq (approved):
It looks ok to me.
Thanks!

joseluisq:
FYI @davlgd I improved your previous work here to be a bit more precise and much easier to use.

The rationale is in the https://github.com/static-web-server/static-web-server/pull/657.

src/text_charset.rs
joseluisq: Resolved by https://github.com/static-web-server/static-web-server/pull/657/changes#diff-5d5cba2eaab0a3e2e4fe5167a48a6ad295a439797a720676eafce0afb11f2593

PR #703, opened 2026-06-20
Always preserve query string on redirects

### LINKED ISSUE #700: Allow preserving query string on redirects

### Feature scope

I have a use case where I'd like to redirect some paths to an external address, while _preserving_ any query strings; i.e. `/some-path?a=query` when redirected needs to send a location header like `Location: http://example.com/some-path?a=query`. As far as I can tell, this is not currently possible because redirects always discard query strings and rewrites cannot have external destinations.

A new option in the `[[advanced.redirects]]` table, e.g. `preserve-query`, which conditionally enables this behavior on a per-redirect basis.

Allowing redirect sources to match the entire URI including query strings would be an alternative solution, but that seems less intuitive and also less backward-compatible.

### Build target

All targets

### Additional context

_No response_

urdh:
## Description
Adds a new preserve-query option to redirects which causes SWS to append any query string received verbatim to the Location header when redirecting. This setting is optional, defaulting to false, which retains existing behavior for backward-compatibility.

## Related Issue
#700

## How Has This Been Tested?
Unit and integration tests have been written and pass (on Arch Linux, amd64, rustc 1.91.1).

### DIFF
--- a/src/redirects.rs
+++ b/src/redirects.rs
@@ -61,7 +61,7 @@ pub(crate) fn pre_process<T>(
         uri_host.push_str(&format!(":{uri_port}"));
     }
     let matched = get_redirection(&uri_host, uri_path, Some(redirects))?;
-    let dest = match replace_placeholders(
+    let mut dest = match replace_placeholders(
         uri_path,
         &matched.source,
         &matched.destination,
@@ -70,6 +70,12 @@ pub(crate) fn pre_process<T>(
         Ok(dest) => dest,
         Err(err) => return handle_error(err, opts, req),
     };
+    if let Some(query) = uri.query() {
+        if !dest.ends_with('?') && !dest.ends_with('&') {
+            dest.push(if dest.contains('?') { '&' } else { '?' });
+        }
+        dest.push_str(query);
+    }
 
     match HeaderValue::from_str(&dest) {
         Ok(loc) => {
@@ -191,6 +197,8 @@ mod tests {
         let r2 = build_placeholder_replacer(&s2);
         let s3 = Regex::new(r"/(prefix/)?(source3)/(.*)").unwrap();
         let r3 = build_placeholder_replacer(&s3);
+        let s4 = Regex::new(r"/source4/(.*)").unwrap();
+        let r4 = build_placeholder_replacer(&s4);
         vec![
             Redirects {
                 host: None,
@@ -212,6 +220,13 @@ mod tests {
                 destination: "/destination3/$2/$3".into(),
                 kind: StatusCode::MOVED_PERMANENTLY,
                 replacer: r3,
+            },
+            Redirects {
+                host: None,
+                source: s4,
+                destination: "/destination4?p=$1".into(),
+                kind: StatusCode::FOUND,
+                replacer: r4,
             },
         ]
     }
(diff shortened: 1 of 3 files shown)


joseluisq (changes requested):
Preserving the query string for redirect is a good idea.
As I wrote in https://github.com/static-web-server/static-web-server/issues/700#issuecomment-4761259551, we should add the query string if available by default to work out of the box.

IMO, this does not represent a "breaking change" per se, but it will enhance what users already have instead.

Here are some adjustments to make:

- Change the `master` branch target to `2.x`. As the `master` now points to the upcoming **v3** (in development). Unless you really wanted your changes to be included in a future v3 release?
- Let's include the query string in the redirect by default if present.
- Let's adjust the tests so we assert this behaviour.
- Revert the new option for redirect query support as we want it built-in.

docs/content/features/url-redirects.md
joseluisq, line 1: The docs have been moved to another repo. You can revert those changes here and make another PR mentioning this change in the new repo. You can find it at https://github.com/static-web-server/docs/tree/master/src/v2.

joseluisq (changes requested):

urdh:
Sounds reasonable!

One thing I hadn't quite thought of when implementing this which might be more important to deal with if this changes default behavior rather than being opt-in is what do to when the _destination_ already has a query string, e.g.:

```toml
[[advanced.redirects]]
source = "/index.html"
destination = "https://static-web-server.net/?src=some-value"
kind = 302
```

Apache has a [QSA](https://httpd.apache.org/docs/2.4/rewrite/flags.html#flag_qsa) option for merging the query strings (the default is to always use the one from _destination_). There's also [QSD](https://httpd.apache.org/docs/2.4/rewrite/flags.html#flag_qsd) for the case where _destination_ **doesn't** have a query string, causing Apache to drop any _source_ query strings (what SWS does now).

I'd lean toward a QSA-like behavior by default, what do you think?

joseluisq:
> Sounds reasonable!
> 
> One thing I hadn't quite thought of when implementing this which might be more important to deal with if this changes default behavior rather than being opt-in is what do to when the _destination_ already has a query string, e.g.:
> 
> ```toml
> [[advanced.redirects]]
> source = "/index.html"
> destination = "https://static-web-server.net/?src=some-value"
> kind = 302
> ```
> 
> Apache has a [QSA](https://httpd.apache.org/docs/2.4/rewrite/flags.html#flag_qsa) option for merging the query strings (the default is to always use the one from _destination_). There's also [QSD](https://httpd.apache.org/docs/2.4/rewrite/flags.html#flag_qsd) for the case where _destination_ **doesn't** have a query string, causing Apache to drop any _source_ query strings (what SWS does now).
> 
> I'd lean toward a QSA-like behavior by default, what do you think?

I believe QSA-like behavior by default is fine too if the query string is present, of course. Let's go with that then.

urdh:
Great! I've updated the PR and re-targeted it to `2.x`. Will try to send a documentation PR later this evening.

docs/content/features/url-redirects.md
urdh: Here's the PR for documentation changes: https://github.com/static-web-server/docs/pull/6.

src/redirects.rs
```diff
@@ -70,6 +70,10 @@ pub(crate) fn pre_process<T>(
         Ok(dest) => dest,
         Err(err) => return handle_error(err, opts, req),
     };
+    if let Some(query) = uri.query() {
+        let delim = if dest.contains('?') { '&' } else { '?' };
+        dest.push_str(&format!("{delim}{query}"));
+    }
```
joseluisq, line 78: This does not look bad, but I would suggest improving it a bit more like appending with correct URL query string semantics.

```suggestion
    if let Some(query) = uri.query() {
        if !dest.ends_with('?') && !dest.ends_with('&') {
            dest.push(if dest.contains('?') { '&' } else { '?' });
        }
        dest.push_str(query);
    }
```

joseluisq (changes requested):
A minor improvement to make.

joseluisq (approved):
Looks ok to me.
Thanks!

PR #713, opened 2026-07-08
Fix: Require authentication for /metrics endpoint when --basic-auth is enabled

T3pp31:
## Summary

Move `basic_auth::pre_process` before `metrics::pre_process` in the request pipeline to ensure that `/metrics` is protected when `--basic-auth` is enabled.

## Problem

Previously, the metrics endpoint was processed before authentication, allowing unauthenticated access to Prometheus metrics when both `--basic-auth` and `--metrics` were enabled. This was reported in GHSA-97q6-jph8-rxgm.

## Solution

Reorder the request pipeline so that basic authentication is checked before the metrics endpoint is processed.

## Testing

- Verified that `/metrics` returns 401 without authentication when `--basic-auth` is enabled
- Verified that `/metrics` returns 200 with valid authentication
- Verified that normal paths still require authentication

## Related

- GHSA-97q6-jph8-rxgm

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -266,12 +266,6 @@ impl RequestHandler {
                     return result;
                 }
 
-                // Metrics endpoint check
-                #[cfg(feature = "metrics")]
-                if let Some(result) = metrics::pre_process(&self.opts, req) {
-                    return result;
-                }
-
                 // CORS
                 if let Some(result) = cors::pre_process(&self.opts, req) {
                     return result;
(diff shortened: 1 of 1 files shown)


joseluisq (changes requested):
This PR targets `master` which belongs to the upcoming v3 (development). What you want I believe is to target `2.x` branch. If so you can switch the target and sync your branch or submit another PR only for v2.

Also, add a couple of tests asserting the basic-auth + metrics combination to make sure that works fine with the new change.

T3pp31:
Since `v2` is now the LTS branch, I've opened #714 targeting `2.x` with the same fix. Closing this PR in favor of #714.

PR #714, opened 2026-07-08
Fix metrics endpoint authentication order

T3pp31:
## Summary
- Move the metrics endpoint check (`metrics::pre_process`) to run **after** the basic authentication check in `RequestHandler::handle`, so that `/metrics` is protected by `--basic-auth` when enabled.
- Previously the metrics pre-processing ran before basic auth, allowing unauthenticated access to the metrics endpoint even when basic auth was configured (auth bypass).
- Add two regression tests in `tests/handler.rs` (gated on `basic-auth` + `metrics` features):
  - `metrics_requires_basic_auth_when_enabled`: requests to `/metrics` without credentials return `401` with a `WWW-Authenticate` header.
  - `metrics_allows_valid_basic_auth_when_enabled`: requests with valid basic auth credentials return `200`.

#### Test plan
- [x] `cargo test --features basic-auth,metrics metrics_requires_basic_auth_when_enabled`
- [x] `cargo test --features basic-auth,metrics metrics_allows_valid_basic_auth_when_enabled`
- [ ] CI (full test suite)

Generated with [Devin](https://devin.ai)

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -257,12 +257,6 @@ impl RequestHandler {
                     return result;
                 }
 
-                // Metrics endpoint check
-                #[cfg(feature = "metrics")]
-                if let Some(result) = metrics::pre_process(&self.opts, req) {
-                    return result;
-                }
-
                 // CORS
                 if let Some(result) = cors::pre_process(&self.opts, req) {
                     return result;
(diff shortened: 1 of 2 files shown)


joseluisq (approved):
Looks fine to me.

Thanks!

PR #722, opened 2026-07-19
Fix pre-compressed variant body truncated to the original file's size

### LINKED ISSUE #721: compression-static: response body truncated to the ORIGINAL file's size (corrupts tiny files whose compressed variant is larger)
### Summary

With `compression-static = true`, when a pre-compressed variant (`.br`/`.gz`/`.zst`) is served, the response **body is truncated to the size of the ORIGINAL (uncompressed) file** instead of the variant's own size.

This goes unnoticed as long as `variant size < original size` (the stream just ends early at EOF). But for **very small files, compressed variants are LARGER than the original** (compression header overhead) — and the variant gets cut off mid-stream. Browsers then decode a partial brotli/gzip stream into a truncated payload, producing errors like:

```
Uncaught SyntaxError: missing binding name           (Firefox, truncated JS module)
Uncaught SyntaxError: "" literal not terminated ...  (import URL cut in half)
```

This is nasty to diagnose in the real world: modern bundlers (Vite/Rollup) routinely emit 30–60 byte JS chunks, and precompression pipelines (e.g. `astro-compressor`) happily generate `.br`/`.gz` variants for them. We chased this for hours across service workers and CDNs before isolating it to SWS. (Bug confirmed on plain binaries, no proxy involved.)

### Reproduction

Affected: **v2.42.0 and v2.43.0** (latest at time of writing), macOS arm64 binaries and Linux containers alike. Minimal config, no proxy.

```bash
mkdir -p repro/site && cd repro

# 39-byte file whose valid .br variant is 43 bytes and .gz variant is 57 bytes
node -e "
const z=require('zlib'),fs=require('fs');
const body=Buffer.from('function n(n){return n}export{n as d};\n');
fs.writeFileSync('site/tiny.js',body);
fs.writeFileSync('site/tiny.js.br',z.brotliCompressSync(body));
fs.writeFileSync('site/tiny.js.gz',z.gzipSync(body,{level:9}));
"

cat > sws.toml <<'EOF'
[general]
host = "127.0.0.1"
port = 8787
root = "./site"
compression = false
compression-static = true
EOF

static-web-server --config-file sws.toml &

curl -s -H 'Accept-Encoding: br' http://127.0.0.1:8787/tiny.js | wc -c
# expected: 43 (the whole .br file)   actual: 39  ← truncated to the ORIGINAL's size

curl -s -H 'Accept-Encoding: br' http://127.0.0.1:8787/tiny.js | brotli -d
# brotli: corrupt input (the stream is incomplete)

curl -s -H 'Accept-Encoding: gzip' http://127.0.0.1:8787/tiny.js | wc -c
# expected: 57   actual: 39
```

Same result with `compression = true` (the dynamic path is fine; only the static-variant path is affected).

### Root cause (v2.43.0)

`compression_static::precompressed_variant()` already stats the variant and returns its metadata in `CompressedFileVariant.metadata` — but the caller drops it:

- `src/static_files.rs` (~L511, and again ~L552):

  ```rust
  .map(|p| (p.file_path, p.encoding));   // <- p.metadata is discarded here
  ```

- `src/static_files.rs` (~L385): the response is then built with the **original** file's metadata while streaming the **variant**:

  ```rust
  let (precomp_path, precomp_encoding) = precompressed_meta;
  ...
  let mut resp = file_reply(headers_opt, file_path, &metadata /* ORIGINAL */, Some(precomp_path), ...)?;
  ```

- `src/response.rs` (~L90): that metadata bounds the streamed body:

  ```rust
  let mut len = meta.len();               // original size, e.g. 39
  ...
  bytes_range(None, len)                  // -> (0, 39): reader budget = 39 bytes
  ```

  so the opened variant file (43 bytes) is streamed but capped at 39 bytes.

Interestingly the code right after already works around half of the symptom — `resp.headers_mut().remove(CONTENT_LENGTH);` — which hides the header mismatch but not the truncated body.

### Suggested fix

Keep the variant's metadata (it is already available, no extra `stat`) and use it for the reply:

```rust
// static_files.rs (~L511/​~L552)
.map(|p| (p.file_path, p.metadata, p.encoding));

// static_files.rs (~L385)
let (precomp_path, precomp_meta, precomp_encoding) = precompressed_meta;
...
let mut resp = file_reply(headers_opt, file_path, &precomp_meta, Some(precomp_path), ...)?;
```

As a bonus, `Content-Length` could then be kept (it would be correct), restoring range support and progress reporting for pre-compressed responses. Using the variant's mtime for conditional headers also seems more correct, since it is the representation actually served.

Happy to submit a PR if that helps.

antoinemichea:
Fixes #721

### Problem

With `compression-static` enabled, a pre-compressed variant response was built with the **original** file's metadata while streaming the **variant** file. `response_body()` bounds the streamed body with `meta.len()`, so any variant **larger** than its original — typical for tiny files, where compression header overhead dominates — was truncated mid-stream. Browsers then decode a partial brotli/gzip/zstd stream, e.g.:

```
Uncaught SyntaxError: missing binding name   (39 of 43 .br bytes served for a 39-byte JS chunk)
```

Full analysis and a standalone reproduction are in #721 (confirmed on v2.42.0 and v2.43.0 release binaries).

### Fix

Honor the contract already documented on `FileMetadata::metadata`:

> Note that if `precompressed_variant` has some value then the `metadata` value will correspond to the `precompressed_variant`.

`compression_static::precompressed_variant()` already stats the variant and returns its metadata — it was simply dropped by the callers. Both resolve sites now expose the variant's metadata (no extra syscall, no public type change).

### Test

Adds `compression_static_variant_larger_than_original`: a 39-byte fixture (`assets/tiny.js`) with a valid 43-byte brotli sibling; the response body must equal the complete variant. The fixtures live under `assets/` so the root directory-listing count fixture (`entries.len() == 9`) is unaffected.

Note: I don't have a local Rust toolchain on this machine, so this relies on CI for build/test validation — the runtime behavior itself was verified against release binaries with the reproduction from #721.


joseluisq (approved):
Good catch!

It looks good to me. However, your changes target the development `v3` (master branch). I believe you want to patch the v2 (stable) which now lives in [2.x](https://github.com/static-web-server/static-web-server/tree/2.x).

I will merge this one and please just create a backport PR targeting the 2.x branch.

PR #734, opened 2026-08-15
Restore recursive matching for `{**}` capture groups

### LINKED ISSUE #692: Regression: brace capture group {**} no longer matches across path separators

- [x] I already searched, and this issue is not a duplicate.

### Issue scope

Regression of existing functionality

### Describe the bug

A brace capture group containing only `**` (e.g. `/app/assets/{**}` → `/assets/$1`) no longer matches across path separators (e.g. `/app/assets/images/logo.png`). It matches only a single path segment, so any nested path 404s. This worked when brace-capture was fixed in #265, and regressed in #506.

### How to reproduce it

```sh
mkdir -p www/assets/sub
echo flat   > www/assets/flat.js
echo nested > www/assets/sub/deep.js

cat > config.toml <<'EOF'
[general]
host = "0.0.0.0"
port = 8787
root = "www"
log-level = "debug"

[advanced]
[[advanced.rewrites]]
source = "/req/{**}"
destination = "/assets/$1"
EOF

static-web-server --config-file config.toml &

curl -s -o /dev/null -w 'flat   -> %{http_code}\n' http://localhost:8787/req/flat.js
curl -s -o /dev/null -w 'nested -> %{http_code}\n' http://localhost:8787/req/sub/deep.js
```

Observed (SWS 2.42.0):

```
flat   -> 200
nested -> 404      # regressed; expected 200
```

### Expected behavior

A capture group around a recursive wildcard should match recursively. `/app/assets/{**}` → `/assets/$1` should serve `assets/a/b/c.js` at any depth, as it did before #506.

In reality:

- `/app/assets/**`     — matches recursively, but provides no capture (`$1` empty, as expected).
- `/app/assets/{**}`   — provides `$1`, but matches only one path segment; nested paths 404.
- The only ways to capture recursively today are non-obvious: `/app/assets/{**/*}`, or moving the slash inside the braces as `/app/assets{/**}` (which also makes `$1` include a leading `/`).

### Complementary information

#265 implemented/fixed rewrite capture groups by repurposing globset's *alternation* syntax: SWS compiles the `source` glob with globset, takes globset's regex translation, and strips `?:` to convert globset's non-capturing `(?:...)` alternation groups into capturing `(...)` groups for `$1` substitution.

At that time `GlobBuilder` was used **without** `literal_separator`, so `{**}` expanded like this:

```
# globset, literal_separator(false)  — pre-#506
/root2/{**}  ->  ^/root2/(?:.*.*)$
```

The double wildcard `.*.*` there is caused by globset treating `**` as two single wildcards when it's not preceded or followed by a `/`. In other words, co-opting the alternation syntax as capture group syntax is breaking globset semantics and behavior.

SWS knows this because it includes `.replace(".*.*", ".*")` in its post-processing, along with stripping `?:`. So after that processing, the example above became `^/root2/(.*)$`, a recursive numbered capture. matching @joseluisq's own description in https://github.com/static-web-server/static-web-server/issues/264#issuecomment-1723094667.

#506 and #501 added `.literal_separator(true)` to `GlobBuilder`. That correctly stops a single `*` from crossing `/`, but because globset is treating the `**` as two single wildcards, it also changed how `{**}` expands:

```
# globset, literal_separator(true)  — post-#506
/root2/{**}  ->  ^/root2/(?:[^/]*[^/]*)$
```

SWS's `.replace(".*.*", ".*")` no longer matches, so the capture stays single-segment. Recursive capture via `{**}` silently broke. There are no tests for the behavior of `{**}`, only `{*}`.

### Build target

Docker linux/arm64

### Environment and specs

- [x] **static-web-server:** 2.42.0 (`joseluisq/static-web-server:2`)
- [x] **Rust:** N/A (repro used the prebuilt official Docker image `joseluisq/static-web-server:2` (SWS 2.42.0), not a local build)
- [x] **OS:** Linux (official SWS image, built `FROM scratch` — no distribution; kernel via Docker)
- [x] **Arch:** arm64
- [x] **Docker:** 29.4.2 (build 055a478)
- [x] **Client:** curl 8.7.1
- [x] globset 0.4.18 (but behavior is governed by the `literal_separator` flag, not the globset version)

#### Evidence

globset regex output for `/root2/{**}`, isolating the `literal_separator` flag added in #506:

| `literal_separator` | globset regex for `/root2/{**}` | recursive? |
| ------------------- | ------------------------------- | ---------- |
| `false` (pre-#506)  | `^/root2/(?:.*.*)$`             | yes        |
| `true`  (post-#506) | `^/root2/(?:[^/]*[^/]*)$`       | **no**     |

SWS's own `--log-level debug` "regex equivalent" output on 2.42.0:

| `source`      | regex produced          | matches nested? | captures `$1`?    |
| ------------- | ------------------------- | --------------- | ----------------- |
| `/req/**`     | `^/req/.*$`               | yes             | **no**            |
| `/req/{**}`   | `^/req/([^/]*[^/]*)$`     | **no**          | yes               |
| `/req{/**}`   | `^/req(/.*)$`             | yes             | yes (leading `/`) |
| `/req/{**/*}` | `^/req/((/?\|.*/)[^/]*)$`  | yes             | yes               |

#### Impact

Breaks the common pass-through pattern `/app/assets/{**}` → `/assets/$1`: it silently 404s any nested asset path. The failure is layout-dependent and silent (flat outputs work, nested ones 404), so it passes casual testing and breaks later.

#### Possible directions

1. Reconcile #506 with #265: keep single `*` non-crossing, but ensure a braced `**` still expands recursively (e.g. special-case `**` inside a capture, or post-process `[^/]*[^/]*` back to `.*` as is already done for `.*.*`).
2. Decouple capture-marking from matching entirely, so capture syntax does not ride on globset alternation and cannot alter `*`/`**` semantics, or use a different library designed to return capture groups
3. At minimum, document that `**` inside `{}` is non-recursive under the current implementation and that recursive capture must be written `{**/*}`.

hdimer:
Fixes #692.

### What regressed

Rewrites and redirects compile the `source` glob with globset, then post-process globset's regex translation as a string: strip `(?-u)`, turn the non-capturing alternation groups `(?:...)` into capturing ones so `$1` works (#265), and collapse `.*.*` into `.*`.

`literal_separator(true)` (#506/#501) changed the shape that last step was written for. globset expands a `**` that isn't a whole path component into two single wildcards, so `{**}` went from `(?:.*.*)` to `(?:[^/]*[^/]*)`. The `.replace(".*.*", ".*")` no longer matches, and `/app/assets/{**}` compiles to `^/app/assets/([^/]*[^/]*)$` - one path segment only. Nested paths 404 silently.

### The fix

One more `.replace()` in each of the two chains, collapsing `([^/]*[^/]*)` back to `(.*)`.

It's scoped to the capture group on purpose. The unscoped form (`[^/]*[^/]*` anywhere) would also widen an *uncaptured* `**` inside a segment, e.g. `/a/**-page`, which would undo the guarantee #506 added. Anchoring on the parens keeps that case at one segment, and the parens can only come from a brace group since globset writes character classes as `[...]`.

### Tests

Three, in the existing fixture style:

- `rewrites_glob_groups_recursive_1` - `/recursive/{**}` serves a nested path.
- `redirects_glob_groups_recursive_1` - same for redirects.
- `redirects_glob_groups_recursive_2{,_literal_separator}` - `/nocross/**-page` still does *not* cross `/`, which is what pins the scoping decision above.

I checked each of them fails on unpatched code, and that the third one fails if the replacement is widened to the unscoped form. `cargo fmt`, `cargo clippy --features=all --tests -- -D warnings` and the full suite are green.

### Two things worth flagging

- This restores the exact `{**}` spelling. A group with more in it, like `{**.js}`, still binds to a single segment - it compiles to `([^/]*[^/]*\.js)`, which the anchor doesn't catch. Deliberately left out to keep the change small; happy to generalize (walk the regex, collapse adjacent `[^/]*` runs only inside a group) if you'd rather fix the class than the case.
- The two post-processing chains are verbatim duplicates ~50 lines apart, and two of the four replaces in them (`?:.*` and `.*.*`) look unreachable under `literal_separator(true)` - which is why this regressed in the first place. I left both alone rather than bury a two-line fix in a refactor. Say the word and I'll extract a helper with unit tests.

The same chain exists verbatim on `2.x`; the patch applies there as-is if you want a backport.

Written with AI assistance; I've verified the behavior and the reasoning myself.

### DIFF
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -546,6 +546,7 @@ impl Settings {
                                 .replace("?:.*", ".*")
                                 .replace("?:", "")
                                 .replace(".*.*", ".*")
+                                .replace("([^/]*[^/]*)", "(.*)")
                                 .to_owned();
                             tracing::debug!(
                                 "url rewrites glob pattern: {}",
@@ -598,6 +599,7 @@ impl Settings {
                                 .replace("?:.*", ".*")
                                 .replace("?:", "")
                                 .replace(".*.*", ".*")
+                                .replace("([^/]*[^/]*)", "(.*)")
                                 .to_owned();
                             tracing::debug!(
                                 "url redirects glob pattern: {}",
--- a/tests/fixtures/toml/redirects.toml
+++ b/tests/fixtures/toml/redirects.toml
@@ -65,6 +65,18 @@ source = "/crop-{[!0-9]}/{*}.{png,gif}"
 destination = "http://localhost/new-crop/$1/$2.$3"
 kind = 301
 
+# Glob groups recursive 1
+[[advanced.redirects]]
+source = "/recursive/{**}"
+destination = "http://localhost/new-recursive/$1"
+kind = 301
+
+# Glob groups recursive 2 (a `**` outside of a group stays within one path segment)
+[[advanced.redirects]]
+source = "/nocross/**-page"
+destination = "http://localhost/new-nocross/"
+kind = 301
+
 # Generic globs need to be at the end
 # Glob groups generic 1
 [[advanced.redirects]]
--- a/tests/fixtures/toml/rewrites.toml
+++ b/tests/fixtures/toml/rewrites.toml
@@ -24,6 +24,11 @@ destination = "/$1x.html"
 source = "/scripts/{*}.{js,mjs}"
 destination = "/assets/$1.$2"
 
+# Glob groups recursive 1
+[[advanced.rewrites]]
+source = "/recursive/{**}"
+destination = "/$1"
+
 # Glob groups 5 (redirect)
 [[advanced.rewrites]]
 source = "**/{*}.{ico}"
--- a/tests/redirects.rs
+++ b/tests/redirects.rs
@@ -339,4 +339,71 @@ pub mod tests {
             }
         };
     }
+
+    #[tokio::test]
+    async fn redirects_glob_groups_recursive_1() {
+        let opts = fixture_settings("toml/redirects.toml");
+        let req_handler_opts = fixture_req_handler_opts(opts.general, opts.advanced);
+        let req_handler = fixture_req_handler(req_handler_opts);
+
+        let mut req = Request::new(());
+        *req.uri_mut() = "http://localhost/recursive/nested/deep/page"
+            .parse()
+            .unwrap();
+
+        let remote_addr = Some(REMOTE_ADDR.parse::<SocketAddr>().unwrap());
+        match req_handler.handle(&mut req, remote_addr).await {
+            Ok(res) => {
+                assert_eq!(res.status(), 301);
+                assert_eq!(
+                    res.headers()["location"],
+                    "http://localhost/new-recursive/nested/deep/page"
+                );
+            }
+            Err(err) => {
+                panic!("unexpected error: {err}")
+            }
+        };
+    }
+
+    #[tokio::test]
+    async fn redirects_glob_groups_recursive_2() {
+        let opts = fixture_settings("toml/redirects.toml");
+        let req_handler_opts = fixture_req_handler_opts(opts.general, opts.advanced);
+        let req_handler = fixture_req_handler(req_handler_opts);
+
+        let mut req = Request::new(());
+        *req.uri_mut() = "http://localhost/nocross/some-page".parse().unwrap();
+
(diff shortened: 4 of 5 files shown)


joseluisq (approved):

joseluisq:
It looks fine to me.
Sure, feel free to backport it to v2.

Thanks!

PR #737, opened 2026-08-17
Suppress fallback page 404 warnings

### LINKED ISSUE #603: Warnings are logged for 404 when using page fallback

### Feature scope

I'm using SWS to serve an application which has a client router, using the `--page-fallback` option.

This works as expected, however I get warnings in the log like this one:
```
2025-12-17 WARN static_web_server::error_page: method=GET uri=/some-client-route status=404 error="Not Found"
```

This clutters the logs and looks to operators as if something is not working as expected, though these 404s are perfectly fine since they are not visible to users.

When using `--page-fallback`, 404 is expected and should not be logged as warning, either completely ignored or logged with info or debug level.

I could manually filter the logs. But I think it would be better to solve this in SWS when using `--page-fallback`.

### Build target

All targets

### Additional context

_No response_

yhuikzdtguioaert:
## Summary

- detect GET 404 responses that the configured page fallback will replace
- suppress the transient 404 warning only for those responses
- keep warnings unchanged for errors that remain visible to clients
- add an integration regression test around the request handler and tracing output

## Verification

- `cargo +1.88.0 fmt --all -- --check`
- `cargo +1.88.0 test --test handler` (6 passed)
- `cargo +1.88.0 test --test handler --no-default-features` (3 passed)
- `cargo +1.88.0 clippy --features all --tests -- -D warnings` after allowing the current baseline-only `duplicated_attributes`, `unnecessary_literal_unwrap`, and `uninlined_format_args` lints
- full all-feature and no-default-feature suites: all relevant tests passed; the existing Windows symlink test failed identically at the base commit

Closes #603

### DIFF
--- a/src/error_page.rs
+++ b/src/error_page.rs
@@ -93,10 +93,36 @@ pub fn error_response(
     page404: &Path,
     page50x: &Path,
 ) -> Result<Response<Body>> {
-    tracing::warn!(
-        method = ?method, uri = ?uri, status = status_code.as_u16(),
-        error = status_code.canonical_reason().unwrap_or_default()
-    );
+    error_response_inner(uri, method, status_code, page404, page50x, true)
+}
+
+/// Builds an error response without emitting an operator-facing warning.
+///
+/// This is used when a later pipeline stage will replace the error response.
+pub(crate) fn error_response_without_logging(
+    uri: &Uri,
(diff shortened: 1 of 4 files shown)


joseluisq (approved):
Thanks!

PR #738, opened 2026-08-18
feat(directory_listing): negotiate listing format via Accept header

### LINKED ISSUE #366: Use Accept header to choose between HTML and JSON directory listings

### Feature scope

As things are now, the decision between HTML and JSON directory listings is made in the configuration. I am unsure about the reasoning behind that, but that’s actually exactly what the `Accept` header is for.

Use content negotiation via the `Accept` header to decide on the directory listing format. There is already `QualityValue` in the tree, it can be reused to parse the `Accept` header. If `application/json` comes up first, the format should be JSON. If `text/html` is found first, it should be HTML. Default if none of the two are present should be HTML.

If necessary, there can also be an explicit `format` query parameter to override the header.

The config setting can be removed then – or ignored with a warning to keep existing configs working.

-

### Build target

All targets

### Additional context

_No response_

aniketch07:
## Summary

Adds a new `auto` value for `--directory-listing-format` (and `DirListFmt::Auto`) that chooses the listing format per request based on the `Accept` header, exactly as suggested in #366:

- `application/json` listed first (q-value order) → JSON listing
- `text/html` listed first → HTML listing
- header absent or neither listed → HTML (fallback)

Existing `html`/`json` values keep working unchanged; the default stays `html`, so current configs are unaffected.

## Implementation notes

- Reuses the existing `Accept`/`QualityValue` types in `src/exts/headers` (the same machinery already used for Accept-Encoding and markdown negotiation) — no new dependencies.
- `Accept::preferred_media_type()` returns the highest-q candidate media type; the mapping to `DirListFmt` lives in `directory_listing/dir.rs`.
- When `Auto` negotiates, the response carries `Vary: Accept` (append-style helper in `exts/http.rs`, same shape as the existing `append_vary_accept_encoding`) so caches don't serve a stale format.

## Verification

- `cargo check --features all` ✅
- `cargo check --no-default-features` ✅ (no feature-leak)
- `cargo clippy --features all -- -D warnings` ✅
- `cargo fmt --all -- --check` ✅
- `cargo test --features all` — full suite green (278 lib tests + integration tests, including 3 new Auto-format integration tests and 5 new unit tests)

Closes #366

### DIFF
--- a/src/directory_listing/autoindex/mod.rs
+++ b/src/directory_listing/autoindex/mod.rs
@@ -39,6 +39,7 @@ pub fn auto_index(opts: DirListOpts<'_>) -> Result<Response<Body>, StatusCode> {
                 is_head: opts.method.is_head(),
                 order_code: opts.dir_listing_order,
                 content_format: opts.dir_listing_format,
+                accept: opts.accept,
                 include_hidden: opts.include_hidden,
                 follow_symlinks: opts.follow_symlinks,
                 #[cfg(feature = "directory-listing-download")]
--- a/src/directory_listing/dir.rs
+++ b/src/directory_listing/dir.rs
@@ -16,6 +16,8 @@ use std::path::Path;
 use crate::body::Body;
 use crate::directory_listing::autoindex::{html_auto_index, json_auto_index};
 use crate::directory_listing::file::{FileEntry, FileType};
+use crate::exts::headers::Accept;
+use crate::exts::http::append_vary_accept;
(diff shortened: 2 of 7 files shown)


aniketch07:
@joseluisq this one is ready for review — implemented the Accept-header negotiation for directory listing, all checks green. Happy to adjust anything you'd like changed.

joseluisq (approved):
Thanks!

